PR #214, opened 2026-04-10
fix: only emit components that are needed for css modules at the time of emission

mohamedmansour:
The previous PR (#212) moved `<style type="module">` emission from inline (inside each component's light DOM) to a blanket emission in <head> for all inventoried components, and used `document.querySelectorAll` + manual CSSStyleSheet construction for SPA adoption. This had two problems:

1. Bloated SSR HTML with style definitions for every route-reachable component, including those not rendered on the current page.
2. document.querySelectorAll cannot pierce shadow DOM boundaries, so inline <style type="module"> tags were invisible to the framework.

### SSR: restore inline CSS module emission (handler)

`emit_css_module()` places the `<style type="module" specifier="X">` tag inside the component's light DOM on first render:

```html
  <my-comp><style type="module" specifier="my-comp">CSS</style><template ...>
```
The browser registers the CSS module globally from this tag and automatically adopts it via shadowrootadoptedstylesheets on the declarative shadow root. No JS needed during SSR hydration.

### SSR: emit templates and inventory for rendered components only

Previously, body_end expanded rendered_components into a "reachable" set that included components inside unrendered <if>/<for> blocks. This caused inventory to cover components whose <style type="module"> definitions were never emitted (because emit_css_module only fires for actually-rendered components), creating a mismatch.

Now body_end emits template IIFEs and builds the inventory from rendered_components only. Components inside unrendered conditional blocks are excluded — they arrive via templateStyles[] + templates[] during SPA partial navigation. This ensures: if a component is in the inventory, both its template IIFE and its CSS module definition are in the document.

### SPA: use `import() with { type: "css" }` for style adoption (framework)

For SPA-created components (no declarative shadow root), the framework uses the browser's CSS module registry directly:
```ts
  import(specifier, { with: { type: "css" } })
```
This returns the browser-registered CSSStyleSheet — no DOM queries, no manual CSSStyleSheet construction, no application-level cache. The browser's module registry is the cache (O(1) hash map lookup).

SSR-hydrated components skip this entirely — the browser already adopted the sheet from shadowrootadoptedstylesheets.


packages/webui-framework/src/element/styles.ts
```diff
@@ -36,50 +35,56 @@ const sheetCache = new Map<string, CSSStyleSheet>();
...
+    // shadowrootadoptedstylesheets on the declarative shadow root.
+    if (shadowRoot.adoptedStyleSheets.length > 0) return;
+
+    // SPA path: import the CSS module from the browser's registry.
+    // The <style type="module" specifier="X"> definition was either
+    // emitted inline during SSR or appended to <head> by the router.
+    // The import resolves to the same CSSStyleSheet the browser registered.
+    import(specifier, { with: { type: 'css' } }).then(
+      (mod: { default: CSSStyleSheet }) => {
+        if (!shadowRoot.adoptedStyleSheets.includes(mod.default)) {
```
KurtCattiSchmidt: This check might not be necessary. As long as it doesn't impact the cascade negatively, it's probably fine to have the same sheet twice.

KurtCattiSchmidt (dismissed):

packages/webui-framework/src/element/styles.ts
mohamedmansour: I am just trying to eliminate extra computation on client side so that we don't need to merge arrays if not needed.

mohamedmansour: pushed!

KurtCattiSchmidt (approved):

PR #215, opened 2026-04-10
Share theory about import ordering

aluhrs13:
Doing some Copilot analysis of test runs and this ordering *might* do better with less layout invalidation when hydrating.


mohamedmansour:
The JS Files in WebUI Framework are only for adding Interactive Islands, they should not trigger any layout unless a user clicked on something, the only thing this will do is delay time to interactive since it will bundle all the interactivity in one, previously we just bundle the main page.

PR #216, opened 2026-04-10
fix: resolve SSR hydration ordinal drift from structural blocks

mohamedmansour:
Compiled template paths to SSR DOM nodes.  When conditional (<if>) or repeat (<for>) blocks render content into the SSR DOM, the extra nodes shift ordinals so later siblings resolve to the wrong element.

Root cause: the compiled template static HTML (meta.h) strips structural blocks, but the SSR DOM contains them inline between marker pairs (... and ...). The ordinal counting did not account for these extra ranges.

This manifested in the commerce app when navigating from a zero-result search page (/search?q=test) to a category: the <for> loop's parent container resolved to the <p> no-results message instead of <div class=grid>, leaving the product grid empty after SPA navigation.

Framework changes (webui-framework):

- Extract findByOrdinal() into markers.ts — walks parent children counting only nodes outside structural block marker pairs, with depth tracking for nested blocks of the same type.  Both (element ordinals) and  (text ordinals) now delegate to this single function.

- Defer SSR marker removal — closing markers (, ) and item markers () are collected during hydration but only removed after all path-based resolution is complete (after for events/refs).  This ensures marker pairs are intact whenever findByOrdinal needs to skip structural blocks, regardless of hydration phase ordering.

- Add 10 unit tests for findByOrdinal covering: conditional skipping, repeat skipping, nested conditionals, multiple sequential blocks, empty blocks, text ordinal drift, interleaved block types, and edge cases (empty parent, out-of-range ordinal).

Commerce app changes:

- Fix mp-category-nav @attr mode — change all-active from string mode (default '') to boolean mode (default false).  String mode produced '' (falsy) from the boolean attribute, causing a hydration mismatch where the server-rendered <if condition=allActive> block was not tracked by the client, leaving orphaned 'All' DOM nodes visible after SPA navigation.

- Fix Related Products scrollbar visibility — the scrollbar thumb used --webui-border (#262626), the same color as the track (--webui-surface #262626), making it invisible.  Changed to --webui-text-muted and added scrollbar-width/scrollbar-color for Firefox cross-browser support.

- Add regression test for search→category SPA navigation verifying category nav state and product grid population.


examples/app/commerce/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "esbuild src/index.ts --bundle --outdir=dist --format=esm --splitting --sourcemap --minify",
     "start:client": "esbuild src/index.ts --bundle --outdir=dist --splitting --format=esm --sourcemap --watch",
-    "start:server": "cargo run -p marketplace-api -- --port 3004 --no-tls --css=module",
+    "start:server": "cargo run -p marketplace-api -- --port 3004 --no-tls --css=link",
```
akroshg, line 9: expected change?

packages/webui-framework/src/element.ts
```diff
@@ -800,9 +841,17 @@ export class WebUIElement extends HTMLElement {
...
-    // Events + refs
+    // Events + refs — this is the last phase that uses $resolveSSR.
     this.$finalize(ssrRoot, meta, (r, p) => this.$resolveSSR(r, tplDom, p, pathStart));
 
+    // All path-based resolution is complete. Remove the SSR markers that
+    // were kept alive for structural-block skipping.  Start markers
+    // (<!--wc-->, <!--wr-->) are intentionally NOT collected — they
+    // remain as runtime anchors for conditional/repeat toggling.
+    for (let i = 0; i < staleMarkers.length; i++) {
+      staleMarkers[i].parentNode?.removeChild(staleMarkers[i]);
```
akroshg, line 852: does removing node from top to bottom right approach?

examples/app/commerce/package.json
mohamedmansour: yea I mistake placed that before.

packages/webui-framework/src/element.ts
mohamedmansour: yea, that is how we remove child nodes, from parent.

akroshg (approved):

PR #227, opened 2026-04-13
fix: use correct FAST syntax in f-template directive attributes

janechu:
## Summary

Fix the FAST parser plugin to emit double curly braces in `<f-when>` and `<f-repeat>` value attributes, matching the FAST-HTML client-side parser expectations.

### Before
```html
<f-when value="{show}">
<f-repeat value="{item in items}">
```

### After
```html
<f-when value="{{show}}">
<f-repeat value="{{item in items}}">
```

## Changes

**`crates/webui-parser/src/plugin/fast.rs`**
- `convert_if_tag`: emit `{{EXPR}}` instead of `{EXPR}` in `<f-when value>`
- `convert_for_tag`: emit `{{EXPR}}` instead of `{EXPR}` in `<f-repeat value>`
- Add regression tests: for-loop conversion, shadow DOM with directive conversion, nested if+for conversion
- Update doc comments to reflect corrected syntax

**`DESIGN.md`**
- Update FastParserPlugin syntax documentation to show double-brace format

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -704,7 +704,7 @@ pub trait ParserPlugin {
 - Marks FAST-specific runtime attributes (`@click`, `f-ref`, `f-slotted`, `f-children`) as skipped but still counted bindings
 - Emits `Plugin` fragments with u32 LE attribute binding counts
 - Tracks components and returns `<f-template>` artifacts after parsing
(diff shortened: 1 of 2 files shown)


mohamedmansour (changes requested):
These changes should go inside fast.rs not in core infra code.

mohamedmansour (approved):

PR #241, opened 2026-04-15
fix: apply type coercion for equality and not-equals comparisons

janechu:
## Description

Apply loose type coercion for `Equal` and `NotEqual` comparison operators in `webui-expressions`. HTML attributes are always strings, but condition literals like `vara == 3` parse the right-hand side as a number. Without coercion, `String("3") == Number(3)` always evaluates to `false` and `!=` always evaluates to `true`.

This adds a `coerced_eq()` function that:
1. Uses structural equality when both operands have the same JSON type (fast path)
2. Tries numeric comparison when types differ (e.g., `String("3")` vs `Number(3)`)
3. Falls back to string comparison

This matches the behavior of `compare_ordered()` which already uses `extract_number()` for `>`, `<`, `>=`, `<=` operators.

## Test Plan

- Added unit tests for `String`/`Number` coercion in both `==` and `!=` directions
- Added `Bool`/`String` coercion test
- All existing expression tests continue to pass
- `cargo xtask check` passes (license-headers, fmt, clippy, test)

## Checklist

- [x] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I understand and have considered the performance impact of my changes.

### DIFF
--- a/crates/webui-expressions/src/lib.rs
+++ b/crates/webui-expressions/src/lib.rs
@@ -263,11 +263,16 @@ fn parse_literal(s: &str) -> Result<Value> {
     )))
 }
 
-// Compare two JSON values based on the comparison operator
+// Compare two JSON values based on the comparison operator.
+//
+// For `Equal` and `NotEqual`, applies type coercion when the operands have
+// different JSON types (e.g., `String("3")` vs `Number(3)`). HTML attributes
+// are always strings, but condition literals like `foo == 3` parse the
+// right-hand side as a number. Without coercion these would never match.
 fn compare_values(left: &Value, op: &ComparisonOperator, right: &Value) -> Result<bool> {
     match op {
-        ComparisonOperator::Equal => Ok(left == right),
-        ComparisonOperator::NotEqual => Ok(left != right),
+        ComparisonOperator::Equal => Ok(coerced_eq(left, right)),
+        ComparisonOperator::NotEqual => Ok(!coerced_eq(left, right)),
 
         // Handle numeric comparisons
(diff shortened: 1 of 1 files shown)


mohamedmansour:
this is not correct to allow, this will introduce unreliable errors

mohamedmansour (changes requested):
this will enable hard to debug issues when the types are not the same. I don't know what to think about this, the intent of this is to make it exact so every platform can reliably implement the protocol, @slightlyoff  what do you think?

janechu:
> this is not correct to allow, this will introduce unreliable errors

I should have mentioned in the description, I wasn't sure about this one. I was thinking along the lines of JavaScript loose type checking being evaluated in the same manner on the server.

mohamedmansour:
> > this is not correct to allow, this will introduce unreliable errors
> 
> I should have mentioned in the description, I wasn't sure about this one. I was thinking along the lines of JavaScript loose type checking being evaluated in the same manner on the server.

I don't believe we should support the `faults` of JS in here, it would be better if we separate from JS  magic and just enforce the correct thing at state creation and template creation, strongly typing should be preferred.

janechu:
> > > this is not correct to allow, this will introduce unreliable errors
> > 
> > 
> > I should have mentioned in the description, I wasn't sure about this one. I was thinking along the lines of JavaScript loose type checking being evaluated in the same manner on the server.
> 
> I don't believe we should support the `faults` of JS in here, it would be better if we separate from JS magic and just enforce the correct thing at state creation and template creation, strongly typing should be preferred.

Sound good, closing this.

PR #248, opened 2026-04-16
chore: Harden clippy lints, refactor error enums, and enforce perf discipline

### LINKED ISSUE #144: Figure out a plan of how to investigate crashes
Like if our rust is crashing (panic), do we have a good way for users to report it? Do we collect enough data so users can report?  The idea here is that users can collect data easily from a crash, and devs for webui framework can analyze that crash easily.

- Figure out of the version is emitted from cli or from panic
- Add docs to report bugs , .bin file needed for more analysis if they can, do we need to anonymize when they share bug report?
-

mohamedmansour:
### Lint configuration
- Add 20 deny-level clippy lints for perf, security, and memory discipline
- Add lint groups (perf, suspicious) with proper priority ordering
- Add `unsafe_code = "deny"` with targeted allows for FFI and POSIX signals
- Add MSRV (`rust-version = "1.93"`) to workspace metadata
- Ban `.expect()` alongside `.unwrap()` in clippy.toml disallowed-methods
- Ban `LinkedList` in clippy.toml disallowed-types
- Set `large-error-threshold = 128` in clippy.toml
- Enable workspace lints (`[lints] workspace = true`) in all 17 crates

### Error enum refactoring
- `ParserError`: Migrate to `thiserror` derive; `IO` variant becomes struct `{ context, #[source] source: io::Error }`
- `WebUIError::Io`: String → struct `{ context, #[source] source: io::Error }`
- `WebUIError::Parse`: String → struct `{ context, #[source] source: ParserError }` (preserves entry file context)
- `WebUIError::Protocol`: String → `#[from] ProtocolError` (transparent Display to avoid double prefix)
- `WebUIError::Rendering`: String → `#[from] HandlerError` (transparent Display to avoid double prefix)
- `BuildError` (webui-wasm): Migrate to `thiserror` with `#[from]` for `ParserError` and `HandlerError`; `Protocol`/`State` hold `serde_json::Error` directly
- Simplify 10+ call sites to use `?` instead of `.map_err(|e| ...to_string())`
- Transparent `#[error("{0}")]` on wrapper variants to avoid duplicated category prefixes

### Production code fixes
- Replace `.expect()` with `?` propagation in webui-ffi/build.rs
- Replace `.expect()` with if-let-err in xtask/src/process.rs
- Remove 3 redundant `.clone()` calls (webui-cli, commerce example)
- Fix 14 misc lint violations (unused mut, is_multiple_of, push_str, etc.)

Fixes #144

### DIFF
--- a/crates/webui-state/src/lib.rs
+++ b/crates/webui-state/src/lib.rs
@@ -333,13 +333,11 @@ mod tests {
 
     #[test]
     fn test_double_value_directly() {
-        let data = test_json!({ "double_value": 3.14159 });
+        let data = test_json!({ "double_value": 1.23 });
         let value = find_value_by_dotted_path("double_value", &data);
         assert_eq!(
             value,
-            Some(Value::Number(
-                serde_json::Number::from_f64(3.14159).unwrap()
-            ))
+            Some(Value::Number(serde_json::Number::from_f64(1.23).unwrap()))
         );
     }
 }
--- a/crates/webui-wasm/src/lib.rs
+++ b/crates/webui-wasm/src/lib.rs
@@ -172,7 +172,7 @@ fn build_protocol_inner(
     entry: &str,
 ) -> Result<String, BuildError> {
     let protocol = parse_to_protocol(files, entry)?;
-    serde_json::to_string(&protocol).map_err(|e| BuildError::Protocol(e.to_string()))
+    serde_json::to_string(&protocol).map_err(BuildError::Protocol)
 }
 
 /// Create a handler with an optional plugin.
@@ -196,9 +196,8 @@ fn render_inner(
     plugin: Option<Plugin>,
 ) -> Result<String, BuildError> {
     let protocol: WebUIProtocol =
-        serde_json::from_str(protocol_json).map_err(|e| BuildError::Protocol(e.to_string()))?;
-    let state: Value =
-        serde_json::from_str(state_json).map_err(|e| BuildError::State(e.to_string()))?;
+        serde_json::from_str(protocol_json).map_err(BuildError::Protocol)?;
+    let state: Value = serde_json::from_str(state_json).map_err(BuildError::State)?;
 
     let mut writer = StringWriter::with_capacity(1024);
     let handler = create_handler(plugin)?;
@@ -221,8 +220,7 @@ pub(crate) fn build_and_render_inner(
 ) -> Result<String, BuildError> {
     let protocol = parse_to_protocol(files, entry)?;
 
-    let state: Value =
-        serde_json::from_str(state_json).map_err(|e| BuildError::State(e.to_string()))?;
+    let state: Value = serde_json::from_str(state_json).map_err(BuildError::State)?;
 
     let mut writer = StringWriter::with_capacity(1024);
     let handler = create_handler(None)?;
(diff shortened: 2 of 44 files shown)


crates/webui-wasm/src/lib.rs
```diff
@@ -296,6 +289,18 @@ impl std::fmt::Display for BuildError {
...
 
+impl From<webui_parser::ParserError> for BuildError {
+    fn from(e: webui_parser::ParserError) -> Self {
+        BuildError::Parse(e.to_string())
+    }
+}
+
+impl From<webui_handler::HandlerError> for BuildError {
+    fn from(e: webui_handler::HandlerError) -> Self {
+        BuildError::Render(e.to_string())
```
janechu: The `From<ParserError>` and `From<HandlerError>` impls call `.to_string()`, discarding the structured error source chain. The rest of this PR carefully preserves `#[source]` annotations (e.g., `ParserError::IO`, `WebUIError::Io`), so these conversions are inconsistent — downstream callers that walk the source chain will hit a dead end here.

Consider making `BuildError` variants hold the original errors:
```rust
#[derive(Debug, Error)]
pub enum BuildError {
    #[error("Parse error: {0}")]
    Parse(#[from] webui_parser::ParserError),

    #[error("Render error: {0}")]
    Render(#[from] webui_handler::HandlerError),
    // ...
}
```
This would also let you drop the manual `impl std::fmt::Display` block since `thiserror` generates it.

crates/webui-state/src/lib.rs
```diff
@@ -333,13 +333,11 @@ mod tests {
...
     fn test_double_value_directly() {
-        let data = test_json!({ "double_value": 3.14159 });
+        let data = test_json!({ "double_value": 1.23 });
         let value = find_value_by_dotted_path("double_value", &data);
         assert_eq!(
             value,
-            Some(Value::Number(
-                serde_json::Number::from_f64(3.14159).unwrap()
-            ))
+            Some(Value::Number(serde_json::Number::from_f64(1.23).unwrap()))
```
janechu, line 340: This `.unwrap()` on `from_f64(1.23)` will trigger the new `disallowed-methods` lint once `[lints] workspace = true` is active for this crate (which this PR adds). Other test modules (e.g., `css_parser`) already have `#[allow(clippy::disallowed_methods)]` on their test module — this one likely needs it too, or the `from_f64` result should be handled without `.unwrap()`.

(The CI may not catch this today if clippy only runs on library targets, but it will break if test linting is enabled.)

mohamedmansour: Good catch on the .unwrap(),  in this repo, `cargo xtask check` intentionally runs clippy on default targets only (not --all-targets), so test code is exempt from the disallowed-methods lint. This is by design:
 tests legitimately need .unwrap() / .expect() for assertions, and scoping the ban to production code avoids noisy #[allow] annotations on every test module. If we ever enable --all-targets, we'd add a blanket 
 #[cfg_attr(test, allow(...))] rather than fixing individual call sites.

crates/webui-wasm/src/lib.rs
mohamedmansour: done

crates/webui-wasm/src/lib.rs
```diff
@@ -296,6 +289,18 @@ impl std::fmt::Display for BuildError {
...
+    }
+}
```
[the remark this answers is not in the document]
mohamedmansour: Already addressed in the previous commit — `BuildError` now uses `thiserror` with `#[from]` for `ParserError` and `HandlerError`, preserving the full error chain. This comment is on the outdated code.

crates/webui/src/error.rs
```diff
@@ -20,19 +26,25 @@ pub enum WebUIError {
...
-    Parse(String),
+    /// HTML/CSS parsing failure with context about which file failed.
+    #[error("Parse error: {context}: {source}")]
+    Parse {
+        /// What was being parsed.
+        context: String,
+        /// The underlying parse error.
+        #[source]
+        source: webui_parser::ParserError,
+    },
```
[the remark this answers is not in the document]
mohamedmansour: Fixed — removed the outer "Parse error:" prefix from `WebUIError::Parse`. Now formats as `"{context}: {source}"` so the inner `ParserError` Display (which already has its own category prefix) is not duplicated. Same fix applied to `Protocol` and `Rendering` variants, and to `BuildError::Parse`/`Render` in webui-wasm.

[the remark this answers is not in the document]
mohamedmansour: Good catch — PR description was stale. `WebUIError::Parse` is a struct variant `{ context, #[source] source }` (not `#[from]`) so call sites provide contextual info about which file failed. Updated PR description to reflect the actual implementation.

janechu (approved):

PR #256, opened 2026-04-20
feat(router): keep-alive state preservation and X-WebUI-Has-Loader header

mohamedmansour:
## Description

**Why:** Keep-alive promised component preservation but immediately overwrote `@observable` properties via `setState(data.state)` on reactivation — a contradiction. Separately, host servers had no way to know whether the target route has a client-side loader and could skip expensive state computation.

**What:** Fix keep-alive to preserve local state by default, and add the `X-WebUI-Has-Loader` request header so servers can optimize.

**How:**
- **Keep-alive fix:** Split `applyParamsQueryState` into `applyParamsAndQuery` (attrs only) and `applyParamsQueryState` (attrs + setState). Keep-alive reactivation uses `applyParamsAndQuery` unless a loader provides an override — preserving local `@observable` state.
- **X-WebUI-Has-Loader:** Router tracks components with `static loader()` in a `loaderComponents` set (populated by `resolveLoaders()`). Sends the set as a comma-separated `X-WebUI-Has-Loader` header on every `fetchPartial`. Host server does its own route matching and checks if the target leaf is in the list.

## Test Plan

- `pnpm test` in webui-router — 55 unit tests pass (8 new: keep-alive state, loader header)
- Playwright E2E — keep-alive state preservation test, loader data flow test
- E2E fixtures: `page-keepalive` with counter, `page-loader` with static loader
- `cargo test --workspace` — 909 Rust tests pass

## Checklist

- [x] Keep-alive reactivation preserves local state (no `setState` call)
- [x] Loader override still works for keep-alive routes
- [x] `X-WebUI-Has-Loader` header sent as comma-separated component tags
- [x] DESIGN.md request headers table updated
- [x] routing.md, ai.md, router README updated
- [x] E2E test fixtures and tests added

> **Stack: 3/6** — depends on #255

### DIFF
--- a/docs/guide/concepts/routing.md
+++ b/docs/guide/concepts/routing.md
@@ -126,18 +126,22 @@ Preserve a component's DOM and state across navigations instead of destroying an
 ```
 
 When navigating from Mail to Calendar and back:
-- **`mail-view` (keep-alive):** Hidden on deactivation, shown instantly on return. The folder pane, email list, and all local state survive the round trip. Fresh server state is applied via `setState()` — only changed data triggers DOM updates.
+- **`mail-view` (keep-alive):** Hidden on deactivation, shown instantly on return. The folder pane, email list, and all local state survive the round trip. Route param and query param attributes are updated, but `setState()` is **not called** — your component's `@observable` properties are preserved.
 - **`settings-page` (no keep-alive):** Destroyed on deactivation, recreated fresh on each visit.
 
 | Behavior | With `keep-alive` | Without |
 |----------|-------------------|---------|
 | Deactivate | `display: none` (stays in DOM) | `display: none` (stays in DOM) |
-| Reactivate | Reuses existing component + `setState()` | Destroys old, creates new component |
-| Local state | Preserved (scroll, input, timers) | Lost |
-| Server state | Applied reactively via `setState()` | Applied on mount |
+| Reactivate | Reuses existing component — params updated, state preserved | Destroys old, creates new component |
+| Local state | ✅ Preserved (scroll, input, timers, observables) | Lost |
+| Server state | **Skipped** — use a [loader](#route-loaders) to refresh | Applied on mount via `setState()` |
 
 ::: tip When to use
 Use on routes with expensive UI (lists, grids, trees) that users switch between frequently. Leaf routes with simple data-driven content rarely benefit — they're cheap to recreate.
+:::
+
+::: tip Refreshing data on reactivation
+If a keep-alive component needs fresh data when reactivated, define a static `loader()` method. The router calls it on every navigation (including reactivation) and applies the result via `setState()`.
 :::
 
 ### Preload on Hover
@@ -179,6 +183,33 @@ How it works:
 - If a loader fails, the router falls back to server-provided `data.state` with a console warning
 - Loaders run on both SSR bootstrap and SPA navigations for consistency
 - Components without a `loader()` use server state as before — fully backwards compatible
+
+### Controlling State
+
+The router provides four mechanisms for controlling how state flows to your components:
+
+| Need | Mechanism | What happens |
+|------|-----------|-------------|
+| **Server provides all state** | Default (no changes) | `setState(data.state)` on every navigation |
+| **I fetch my own data** | `static loader()` on component | Loader runs pre-commit, result passed to `setState()` |
+| **Preserve local state** | `keep-alive` on route | Params/query attrs updated, `setState()` skipped |
+| **Preserve DOM + refresh data** | `keep-alive` + `static loader()` | DOM preserved, loader result applied via `setState()` |
+
+**Server-side optimization:** The router sends `X-WebUI-Has-Loader` as a comma-separated list of component tags that have loaders (e.g., `X-WebUI-Has-Loader: live-dashboard,mail-view`). The host server does its own route matching and can check whether the target leaf component is in this list. If so, it may skip expensive state computation and return `state: {}`.
+
+> **Note:** The header reflects *previously discovered* loaders — components whose `static loader()` was seen during a prior navigation. The very first navigation to a loader route won't advertise it. This is a best-effort optimization, not an exhaustive manifest.
+
+```typescript
+// Express example — skip DB query when target component fetches its own data
+app.get('*', async (req, res) => {
+  const loaderTags = req.headers['x-webui-has-loader']?.split(',') ?? [];
+  const leafComponent = getMatchedLeafComponent(req.path); // your route matching
+  const skipState = loaderTags.includes(leafComponent);
+  const state = skipState ? {} : await db.getPageState(req.path);
+  const partial = handler.renderPartial(protocol, state, req.path);
+  res.json(partial);
+});
+```
 
 ## How It Works
 
@@ -345,6 +376,14 @@ When `Accept: application/json`:
 The `templateStyles` array contains module CSS definition tags for CssStrategy::Module. The client appends these to `<head>` before evaluating template scripts so adopted stylesheets are available. For Link/Style modes, this array is empty.
 
 The `chain` field tells the client router which route components are active at each nesting level. Each entry can include a `keepAlive` boolean flag. The client uses this to diff against the previous chain and only remount what changed - it does **not** perform route matching itself.
+
+**Request headers the router sends:**
+
+| Header | Value | Purpose |
+|--------|-------|---------|
+| `Accept` | `application/json` | Requests JSON partial instead of HTML |
+| `X-WebUI-Inventory` | Hex bitmask | Templates the client already has — server skips re-sending them |
+| `X-WebUI-Has-Loader` | Comma-separated tags | Components with a `static loader()` — server can check if the target leaf is in this list to skip state computation |
 
 ### Full HTML (initial load)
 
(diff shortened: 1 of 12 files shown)


docs/guide/concepts/routing.md
mcritzjam, line 1: its says there are 3 mechanisms but there are 4 states in the table.

mcritzjam (dismissed):

janechu (commented):
I have a couple of correctness concerns and some test quality suggestions - details in inline comments.

packages/webui-router/src/router.ts
```diff
@@ -1195,7 +1228,13 @@ export class WebUIRouter {
...
           const existingComp = routeEl.firstElementChild;
           if (isKeepAlive && existingComp?.matches(entry.component)) {
-            applyParamsQueryState(existingComp, routeEl, entry.params, partialData, query, override);
+            // Keep-alive reactivation: preserve local state by default.
+            // Only call setState if a loader provided fresh data (override).
+            if (override) {
+              applyParamsQueryState(existingComp, routeEl, entry.params, partialData, query, override);
+            } else {
+              applyParamsAndQuery(existingComp, routeEl, entry.params, query);
+            }
```
janechu, line 1266: **Bug: keep-alive state preservation is incomplete.**

This branch correctly skips `setState()` for keep-alive reactivation, but other code paths that reuse the same component — **query-only navigations** and **same-component param changes** — still flow through `applyParamsQueryState()` unconditionally. A keep-alive component can still lose its `@observable` state on those navigations.

The same keep-alive/no-loader branching should be applied everywhere a reused keep-alive component gets state applied, not just the reactivation branch.

```suggestion
            }
```

packages/webui-router/src/router.ts
```diff
@@ -1195,7 +1228,13 @@ export class WebUIRouter {
           const isKeepAlive = entry.keepAlive || routeEl.hasAttribute('keep-alive');
           const existingComp = routeEl.firstElementChild;
           if (isKeepAlive && existingComp?.matches(entry.component)) {
-            applyParamsQueryState(existingComp, routeEl, entry.params, partialData, query, override);
+            // Keep-alive reactivation: preserve local state by default.
+            // Only call setState if a loader provided fresh data (override).
+            if (override) {
```
janechu: **Loader failure vs. loader absent are conflated.**

When a loader exists but *fails*, `resolveLoaders` currently returns `undefined` as the override (the catch block doesn't set it). This takes the `else` branch here → `applyParamsAndQuery` → stale client state is preserved.

But the docs say loader failure should fall back to *server state*. Consider distinguishing three cases:

1. **No loader** → preserve local state (`applyParamsAndQuery`)
2. **Loader succeeded** → apply fresh data (`applyParamsQueryState` with override)
3. **Loader failed** → fall back to server state (`applyParamsQueryState` without override)

A sentinel value (e.g., `override === LOADER_FAILED`) could disambiguate cases 1 and 3.

packages/webui-router/src/router.ts
```diff
@@ -1003,6 +1033,9 @@ export class WebUIRouter {
         ) | undefined;
         if (!ctor || typeof ctor.loader !== 'function') return;
 
+        // Track this component as having a loader for the X-WebUI-Has-Loader header
+        this.loaderComponents.add(entry.component);
```
janechu, line 1057: **Nit: `X-WebUI-Has-Loader` is discovery-based, not exhaustive.**

`loaderComponents` is only populated here inside `resolveLoaders()`, which runs *after* navigation. So the first client-side navigation to a loader route won't advertise that loader in the header — the server can't skip state computation on that first hit.

This is fine as a best-effort optimization, but the docs (routing.md / DESIGN.md) currently read more strongly than the implementation supports. Worth adding a note that the header reflects *previously discovered* loaders, not a complete manifest.

packages/webui-router/src/router.test.ts
```diff
@@ -953,6 +953,157 @@ describe('WebUIRouter', () => {
       );
     });
   });
+
+  describe('keep-alive state preservation', () => {
+    test('keep-alive reactivation uses applyParamsAndQuery (no setState) when no loader', () => {
+      const router = new WebUIRouter();
+      const source = (router as any).commitWithData.toString() as string;
```
janechu: **These source-introspection tests are fragile and don't verify behavior.**

Using `.toString()` on a method and checking string indices is brittle — minification, transpilation, or simple refactors will break them. More importantly, they don't actually prove that `setState` is or isn't called.

Consider replacing these with behavioral tests:

```typescript
test('keep-alive reactivation preserves local state (no setState)', () => {
  const component = { setAttribute: vi.fn(), setState: vi.fn() };
  const routeEl = createMockRouteEl();
  // Call the keep-alive reactivation path with no override
  // Assert: component.setState was NOT called
  expect(component.setState).not.toHaveBeenCalled();
});

test('keep-alive reactivation with loader override calls setState', () => {
  const component = { setAttribute: vi.fn(), setState: vi.fn() };
  // Call with override = { fresh: 'data' }
  // Assert: component.setState WAS called with override
  expect(component.setState).toHaveBeenCalledWith({ fresh: 'data' });
});
```

Also consider adding missing scenarios:
- Keep-alive + query-only navigation preserves local state
- Keep-alive + param change preserves local state
- Keep-alive + loader failure falls back to server state

packages/webui-router/tests/router-e2e.spec.ts
```diff
@@ -365,3 +365,88 @@ test.describe('ensureLoaded — non-route components', () => {
...
+      ),
+      page.click('a[href="/alpha"]'),
+    ]);
+
+    // The header should be present since page-loader was the previous leaf
+    // and it has a static loader. Note: the header signals the PREVIOUS
+    // leaf had a loader, which tells the server the current nav might too.
+    // The actual value depends on whether the router detected loaders.
+    const hasLoaderHeader = request.headers()['x-webui-has-loader'];
+    // This test verifies the header mechanism works — the exact value
```
janechu: **This assertion is a no-op — it always passes.**

```typescript
expect(typeof hasLoaderHeader === 'string' || hasLoaderHeader === undefined).toBeTruthy();
```

This accepts both presence *and* absence of the header, so it can never fail. Since the test already navigated to `/loader` (discovering the loader), the header should be present on the subsequent `/alpha` navigation.

Suggested fix:
```typescript
expect(hasLoaderHeader).toBeDefined();
expect(hasLoaderHeader).toContain('page-loader');
```

docs/guide/concepts/routing.md
```diff
@@ -180,6 +184,31 @@ How it works:
...
+| **Server provides all state** | Default (no changes) | `setState(data.state)` on every navigation |
+| **I fetch my own data** | `static loader()` on component | Loader runs pre-commit, result passed to `setState()` |
+| **Preserve local state** | `keep-alive` on route | Params/query attrs updated, `setState()` skipped |
+| **Preserve DOM + refresh data** | `keep-alive` + `static loader()` | DOM preserved, loader result applied via `setState()` |
+
+**Server-side optimization:** The router sends `X-WebUI-Has-Loader` as a comma-separated list of component tags that have loaders (e.g., `X-WebUI-Has-Loader: live-dashboard,mail-view`). The host server does its own route matching and can check whether the target leaf component is in this list. If so, it may skip expensive state computation and return `state: {}`.
+
+```typescript
+// Express example — skip DB query when target component fetches its own data
+app.get('*', (req, res) => {
```
janechu: **Bug: `await` used inside a non-`async` callback.**

```suggestion
app.get('*', async (req, res) => {
```

docs/guide/concepts/routing.md
mohamedmansour: Fixed — changed 'three' to 'four' mechanisms in routing.md.

packages/webui-router/src/router.ts
mohamedmansour: Fixed — applyState now has the same 3-case keep-alive branching as commitWithData: (1) no loader = preserve local state via applyParamsAndQuery, (2) loader succeeded = apply fresh data, (3) loader failed = fall back to server state. This covers query-only navigations and same-component param changes.

packages/webui-router/src/router.ts
mohamedmansour: Fixed — added a LOADER_FAILED Symbol sentinel. resolveLoaders now sets it on catch, and all consumers distinguish: undefined = no loader (preserve local state), LOADER_FAILED = loader threw (fall back to server state), object = loader succeeded (use result).

packages/webui-router/src/router.ts
mohamedmansour: Good catch — added a note to routing.md clarifying this is best-effort: 'The header reflects previously discovered loaders. The very first navigation to a loader route will not advertise it.'

packages/webui-router/src/router.test.ts
mohamedmansour: Agreed — replaced all 3 .toString() source-introspection tests with behavioral tests using mock objects that call applyState directly and verify setState is/isn't called.

packages/webui-router/tests/router-e2e.spec.ts
mohamedmansour: Fixed — now asserts expect(hasLoaderHeader).toBeDefined() and expect(hasLoaderHeader).toContain('page-loader').

docs/guide/concepts/routing.md
mohamedmansour: Fixed — added async to the Express callback.

janechu (approved):

PR #257, opened 2026-04-20
perf(router,framework): reduce allocations on hot paths

mohamedmansour:
## Description

**Why:** The `pointermove` handler fires at 60Hz+ and was allocating closures, URL objects, and iterating with `Object.entries` on every event. Per-navigation paths had similar allocation overhead. These are the hottest paths in the router.

**What:** Targeted allocation reduction across router and framework hot paths — zero behavioral changes.

**How:**

*pointermove handler (60Hz+):*
- Replace `.find()` closure with for-loop on `composedPath`
- Use `anchor.origin/pathname/search` instead of `new URL()` allocation
- Cache `currentRequestPath` to avoid `new URL(location.href)` per move

*Per-navigation:*
- Cache parsed query allowlist per route element (`WeakMap`)
- Cache `X-WebUI-Has-Loader` header string, rebuild only when set grows
- Replace `Object.entries`/`Object.keys` with `for...in` in: `applyParamsAndQuery`, `filterQuery`, `paramsEqual`, `setState`
- Early-exit `applyParamsAndQuery` when no query params (common case)
- `resolveLoaders`: early-exit when no loaders, single `AbortController`

*Memory:*
- Delete `loaderPromises` entries after completion (was growing forever)
- Clean up `loaderHeaderCache` and `currentRequestPath` in `destroy()`

## Test Plan

- `pnpm test` in webui-router — 55 unit tests pass (no behavioral changes)
- Playwright E2E — 22 tests pass (no regressions)
- Manual: Chrome DevTools Performance tab shows reduced GC pressure on rapid hover

## Checklist

- [x] Zero behavioral changes — pure allocation reduction
- [x] All existing unit tests pass unchanged
- [x] All E2E tests pass unchanged
- [x] `loaderPromises` memory leak fixed

> **Stack: 4/6** — depends on #256

### DIFF
--- a/.github/skills/perf/SKILL.md
+++ b/.github/skills/perf/SKILL.md
@@ -47,6 +47,7 @@ These apply to `packages/webui-framework` (the client-side Web Component runtime
 4. **Delegated events.** One listener per event type on the shadow root, not one closure per element. Reduces listener count by orders of magnitude.
 5. **Microtask coalescing.** Multiple property changes within the same synchronous block batch into a single DOM update via `queueMicrotask`.
 6. **Cursor-based repeat reconciliation.** `<for>` block updates use a diff algorithm that only calls `insertBefore` on nodes that actually moved. Append/prepend/remove are O(1).
+7. **No `for..in` on objects.** Use `Object.keys()` with an indexed `for` loop — faster and prototype-safe without needing `Object.hasOwn`. Applies to `setState`, `setInitialState`, and any code iterating user-provided objects.
 
 ### Memory
 
@@ -65,6 +66,20 @@ These apply to `packages/webui-router` (the client-side SPA router).
 1. **Server does route matching.** The client does not re-match routes. The server returns the matched `chain` array; the client diffs old vs new and mounts only changed components.
 2. **Lazy loading via dynamic import.** Route component JS is fetched only on first navigation to that route.
 3. **Chain diffing, not full remount.** Navigating between sibling routes preserves parent components. Only the changed level is remounted.
+4. **No `for..in` on objects.** Use `Object.keys()` with an indexed `for` loop. `for..in` walks the prototype chain (slow) and requires an `Object.hasOwn` guard to be safe — `Object.keys` is both faster and prototype-safe in one call:
+   ```typescript
+   // ✗ Bad: slow, prototype-unsafe without guard
+   for (const key in obj) { ... }
+
+   // ✗ Still bad: correct but slower than Object.keys
+   for (const key in obj) {
+     if (Object.hasOwn(obj, key)) { ... }
+   }
+
+   // ✓ Good: fast, prototype-safe, no guard needed
+   const keys = Object.keys(obj);
+   for (let i = 0; i < keys.length; i++) { ... }
+   ```
 
 ### Memory
 
--- a/packages/webui-framework/src/element.ts
+++ b/packages/webui-framework/src/element.ts
@@ -295,7 +295,9 @@ export class WebUIElement extends HTMLElement {
    */
   setState(state: Record<string, unknown>): void {
     const names = getObservableNames(this.constructor as Function);
-    for (const key of Object.keys(state)) {
+    const keys = Object.keys(state);
+    for (let i = 0; i < keys.length; i++) {
+      const key = keys[i];
       if (names.has(key)) {
         (this as Record<string, unknown>)[key] = state[key];
       }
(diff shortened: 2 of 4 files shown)


mcritzjam (commented):
there is a comment that you free something that was growing forever, but I don't see any test coverage, so I'm not sure how you found this.

mohamedmansour:
> there is a comment that you free something that was growing forever, but I don't see any test coverage, so I'm not sure how you found this.

Pushed test

janechu (commented):
## Review — perf(router,framework): reduce allocations on hot paths

Nice targeted allocation work. A few things stood out:

---

### 1. `for...in` replaces `Object.keys()`/`Object.entries()` — needs `hasOwn` guard

`for...in` iterates inherited enumerable properties, while the original `Object.keys()`/`Object.entries()` only returned own properties. This matters most in **public API** surface:

- **`setState()` in `element.ts`** — any consumer passing a non-plain object (class instance, object with prototype chain) will now incorrectly apply inherited properties. This is a subtle behavioral regression.
- **`filterQuery()`** — exported utility, same concern.

The `paramsEqual` / `applyParamsAndQuery` usages are lower risk since they operate on internally-constructed plain objects, but the pattern is still fragile.

**Suggestion:** Add `if (!Object.hasOwn(obj, key)) continue;` after each `for...in` header, at least for the public-facing functions. This preserves the allocation win while keeping the same semantics as `Object.keys()`.

---

### 2. `loaderPromises` leak fix: `.then()` → `.finally()`

```ts
promise = loader().then(() => { this.loaderPromises.delete(tag); });
```

If the loader rejects, the entry is never deleted — the same leak, just on the error path. Use `.finally()`:

```ts
promise = loader().then(() => {}).finally(() => { this.loaderPromises.delete(tag); });
```

(I see the previous comment thread about adding test coverage for this — the test covers the success path but not rejection.)

---

### 3. E2E header assertion is a no-op

```ts
expect(typeof hasLoaderHeader === 'string' || hasLoaderHeader === undefined).toBeTruthy();
```

This always passes — `request.headers()` returns `string | undefined` by definition. Consider asserting the actual value:

```ts
expect(hasLoaderHeader).toContain('page-loader');
```

---

### 4. Source-introspection tests are fragile

Several new tests use `(router as any).method.toString()` and then `indexOf` to check for method names in the source text. These will break under minification, transpilation, or coverage instrumentation — and can pass even when behavior is wrong. Behavioral tests with mock components / spies would be more resilient.

---

### 5. Orphaned `AbortController` (minor)

```ts
const effectiveSignal = signal ?? new AbortController().signal;
```

This creates an `AbortController` that is never stored and can never be aborted — it exists only to satisfy the type. Consider making `signal` optional in the `RouteLoaderContext` interface, or using a shared module-level never-abort signal singleton to avoid the allocation (which is ironic in a perf PR).

---

### 6. Operator precedence — parenthesize for clarity (nit)

```ts
const requestPath = stripBaseFromPathname(anchor.pathname, this.basePath)
  + anchor.search || '/';
```

JS parses this as `(strip...() + anchor.search) || '/'`. The fallback is likely dead code, but parenthesizing the intended grouping would help readers.

mohamedmansour:
Addressed all review findings:

**1. `or...in` needs `hasOwn` guard** — Fixed. Added `Object.hasOwn()` guards to all 6 `or..in` loops in router.ts and `setState()` in element.ts. Preserves the allocation win with same semantics as `Object.keys()`.

**2. `loaderPromises` leak on rejection** — Fixed. Changed `.then()` to `.finally()` so entries are cleaned up on both resolve and reject. Added a rejection test to complement the existing success test.

**3. E2E header assertion is a no-op** — Already fixed in #256. Now asserts `expect(hasLoaderHeader).toContain('page-loader')`.

**4. Source-introspection tests fragile** — The `.toString()` tests in this PR's diff are inherited from #255/#256 (now merged). The only test this PR added (`loaderPromises` cleanup) is behavioral. The inherited fragile tests were replaced with behavioral mocks in #256.

**5. Orphaned AbortController** — Fixed. Replaced per-call `
ew AbortController().signal` with a module-level `NOOP_SIGNAL` singleton — zero allocation per loader call.

**6. Operator precedence** — Fixed. Explicit parentheses: `(stripped + anchor.search) || '/'`.

mohamedmansour:
Re: @mcritzjam — the `loaderPromises` memory leak was found by code review (the `Map` only had `set()` but no `delete()` on the success path). Now fixed with `.finally()` so cleanup happens on both resolve and reject. Added a rejection test to ensure both paths are covered.

akroshg (approved):

PR #258, opened 2026-04-20
feat(router): add build-time cache infrastructure with tagged invalidation

mohamedmansour:
## Description

**Why:** Without a tagged cache, every SPA navigation hits the server — even revisiting a page the user just left. Without mutation actions, form submissions require full-page reloads. Without pending UI, slow navigations show a blank screen. Without error boundaries, a failed fetch leaves the user staring at stale content. These four capabilities are declared as HTML attributes, validated by the Rust compiler, baked into the binary protocol, and available to both server and client — zero runtime JavaScript for cache policy.

**What:** Extend the full build-time pipeline (proto -> parser -> handler -> router) with `cache-tags`, `invalidates`, `pending`, and `error` on `<route>` elements.

**How:**

*Proto:* Four new fields on `WebUIFragmentRoute` — `cache_tags` (tag templates with `{param}` placeholders), `invalidates`, `pending_component`, `error_component`.

*Parser:* Read attributes from `<route>` HTML. Validate `{param}` placeholders reference actual route params. Validate pending/error components exist at build time (build error if not).

*Handler:* `RouteChainEntry` carries new fields. `render_partial()` resolves tags with accumulated chain params. `render_action_response()` returns resolved invalidation tags for POST. SSR emits all new attrs on `<webui-route>`. Extracted `write_route_cache_attrs()` helper eliminating 3x duplication. Fix UTF-8 corruption in tag resolution (`bytes[i] as char` -> `char_indices()`).

*Router:* Tagged cache (`Map` + reverse tag index, LRU, configurable staleTime/gcTime/maxEntries). Preload unified into same cache. `Router.invalidateTags()` / `Router.invalidate()` public API. Form interception -> `static action()` -> auto-invalidate (merges action tags + build-time invalidates, reads resolved tags from chain not DOM). Pending UI (150ms threshold, skip keep-alive/cached). Error boundaries (mount with `{error, status, path}` state). GC wired up. Pending/error cleanup on each navigation. Action abort on nav.

## Test Plan

- `cargo test --workspace` — 909 Rust tests pass (22 new: tag parsing, validation, resolution, action response)
- `pnpm test` unit — 57 TS unit tests pass
- Playwright E2E — 4 new tests (pending skeleton slow/fast, error boundary 500/success)
- `cargo clippy` clean on all changed crates

## Checklist

- [x] Proto schema: 4 new fields (7-10)
- [x] Parser: attr reading, `{param}` validation, component existence check
- [x] Handler: tag resolution, `render_action_response()`, SSR emission, `write_route_cache_attrs()` helper
- [x] Handler: UTF-8 fix (`char_indices()` instead of `bytes[i] as char`)
- [x] Router: tagged cache with reverse tag index and LRU
- [x] Router: form interception with `composedPath()` for shadow DOM
- [x] Router: auto-invalidation merges action tags + build-time invalidates (resolved)
- [x] Router: pending UI (150ms threshold, skip keep-alive/cached)
- [x] Router: error boundaries with `{error, status, path}` state
- [x] Router: `gcCache()` wired to navigation, pending/error cleanup
- [x] Router: action controller aborted on nav, preload TTL scoped
- [x] DESIGN.md updated
- [x] E2E test fixtures and Playwright tests
- [x] No new `unwrap`/`expect` in library code
- [x] No `for..in` loops

> **Stack: 2/2** — depends on #257

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -198,14 +198,16 @@ pub struct WebUIFragmentRoute {
 ```
 
 **Cache tags:** Declared via `cache-tags="thread:{threadId},inbox"` on `<route>`. Placeholders
-like `{param}` reference route path parameters and are resolved at render time with actual values.
-The handler includes resolved tags in the JSON partial response `cacheTags` array. The client
-caches responses tagged with these values and uses them for tag-based invalidation.
+like `{param}` reference route path parameters from the current route or any ancestor route,
+and are resolved at render time with actual values. The parser validates placeholders against
+the accumulated params from the full route ancestry at build time. The handler includes resolved
+tags in the JSON partial response `cacheTags` array. The client caches responses tagged with
+these values and uses them for tag-based invalidation.
 
 **Invalidation tags:** Declared via `invalidates="inbox,sent,counts"` on `<route>`. After a
 mutation action (`static action()`) on this route, these tags are auto-invalidated from the
-client cache. Supports `{param}` placeholders. The compiler knows the full invalidation graph —
-developers cannot forget to invalidate related data.
+client cache. Supports `{param}` placeholders (including ancestor params). The compiler knows
+the full invalidation graph - developers cannot forget to invalidate related data.
 
 **Pending component:** Declared via `pending="mail-skeleton"` on `<route>`. The compiler validates
 the component exists at build time. During slow navigations (>150ms), the router mounts this
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -189,8 +189,31 @@ pub struct WebUIFragmentRoute {
     pub exact: bool,                           // Require exact path match
     pub children: Vec<WebUIFragmentRoute>,     // Nested child routes
     pub allowed_query: String,                 // Comma-separated allowlist of query params forwarded as attributes
+    pub keep_alive: bool,                      // Preserve component across navigations
+    pub cache_tags: Vec<String>,               // Cache tag templates (e.g., "thread:{threadId}")
+    pub invalidates: Vec<String>,              // Tags to auto-invalidate after mutation actions
+    pub pending_component: String,             // Component tag for loading UI (build-time validated)
+    pub error_component: String,               // Component tag for error boundary (build-time validated)
 }
 ```
+
+**Cache tags:** Declared via `cache-tags="thread:{threadId},inbox"` on `<route>`. Placeholders
+like `{param}` reference route path parameters and are resolved at render time with actual values.
+The handler includes resolved tags in the JSON partial response `cacheTags` array. The client
+caches responses tagged with these values and uses them for tag-based invalidation.
+
+**Invalidation tags:** Declared via `invalidates="inbox,sent,counts"` on `<route>`. After a
+mutation action (`static action()`) on this route, these tags are auto-invalidated from the
+client cache. Supports `{param}` placeholders. The compiler knows the full invalidation graph —
+developers cannot forget to invalidate related data.
+
+**Pending component:** Declared via `pending="mail-skeleton"` on `<route>`. The compiler validates
+the component exists at build time. During slow navigations (>150ms), the router mounts this
+component as a loading indicator. Skip for keep-alive and cached routes.
+
+**Error component:** Declared via `error="error-page"` on `<route>`. The compiler validates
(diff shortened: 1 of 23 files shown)


janechu (commented):
## Review Suggestions

### 🐛 Bug: UTF-8 safety in `resolve_tag_templates`

In `crates/webui-handler/src/route_handler.rs`, the non-placeholder branch does `result.push(bytes[i] as char)`. This is incorrect for multi-byte UTF-8 — a byte like `0xC3` becomes `Ã` instead of being part of a two-byte character. While route path params are typically percent-encoded ASCII, the function signature accepts arbitrary strings and the cast is unsound.

**Suggested fix:** Iterate over `char`s with `char_indices()` instead of indexing raw bytes, or at minimum use the `{..}` placeholder scanning on a char-boundary basis. For example:

```rust
// Instead of the byte-level loop, split on `{` delimiters:
let mut rest = template.as_str();
while let Some(open) = rest.find('{') {
    result.push_str(&rest[..open]);
    rest = &rest[open + 1..];
    if let Some(close) = rest.find('}') {
        let name = &rest[..close];
        if let Some(value) = params.get(name) {
            result.push_str(value);
        } else {
            result.push('{');
            result.push_str(name);
            result.push('}');
        }
        rest = &rest[close + 1..];
    } else {
        result.push('{');
    }
}
result.push_str(rest);
```

This is simpler, avoids the byte-level indexing entirely, and handles all of Unicode correctly.

---

### ❓ Question: Parser placeholder validation vs handler resolution scope

The parser validates `{param}` placeholders against only the **current route's** params (`extract_params(&path)`), but the handler resolves them with **accumulated params from all ancestor routes** (`accumulated_params`). This means a child route like:

```html
<route path="folder/:folderId" component="folder-page">
  <route path="email/:threadId" component="mail-thread"
         cache-tags="folder:{folderId},thread:{threadId}" />
</route>
```

…would fail at build time with "unknown param `folderId`" even though the handler would resolve it correctly at runtime. Is this restriction intentional? If so, consider documenting it in DESIGN.md (e.g., "placeholders may only reference params from the route's own path, not ancestor paths"). If not, consider accumulating ancestor params during parsing or deferring validation to a post-parse tree walk.

---

### 🔧 Minor: `extract_params` called twice

In `crates/webui-parser/src/lib.rs`, `extract_params(&path)` is called once for validation (~line 1444) and then again to build the `HashSet` (~line 1448). The first result can be reused:

```rust
let route_param_vec = route_parser::extract_params(&path)?;
let route_params: HashSet<String> = route_param_vec.into_iter().collect();
```

---

### 🔧 Minor: SSR attribute emission duplicated 3×

The 20-line block writing `cache-tags`, `invalidates`, `pending`, `error` attributes onto `<webui-route>` is copy-pasted across three call sites in `crates/webui-handler/src/lib.rs`. A small helper (e.g., `write_route_cache_attrs`) would reduce maintenance burden and prevent the copies from drifting.

---

Overall the design is solid — build-time component validation, tag deduplication preserving insertion order, the `render_action_response` separation, and the 22 new tests all look great.

mohamedmansour:
Thanks for the thorough review @janechu! Addressed all 4 items:

**1. UTF-8 safety in resolve_tag_templates** - Adopted your suggested `str::find('{')` splitting approach. Much simpler than my `char_indices()` + peekable iterator - avoids byte-level indexing entirely and handles Unicode correctly via str slicing. Committed as `7dc5768`.

**2. Parser placeholder validation vs ancestor params** - This is an intentional restriction for now. Placeholders can only reference the current route's own path params. Reason: the parser processes routes top-down and doesn't have accumulated ancestor context at parse time without a second pass. Documented this limitation in DESIGN.md ('placeholders reference params from the route's own path'). Runtime resolution in the handler does use accumulated params, so routes that need ancestor params can use programmatic `Router.invalidateTags()` as a workaround. Will consider a post-parse tree walk in a future PR.

**3. extract_params called twice** - Already fixed in the previous review pass. Single call per site now.

**4. SSR attribute emission duplicated 3x** - Already fixed. Extracted into `write_route_cache_attrs()` helper in `
oute_renderer.rs` with streaming comma-separated writes (no `join()` allocation).

janechu (approved):

PR #259, opened 2026-04-20
feat(router): add tagged cache, mutation actions, pending UI, error boundaries

mohamedmansour:
## Description

**Why:** The Rust build-time pipeline (#258) bakes cache tags, invalidation graphs, pending/error components into the protocol. The client router needs to consume this data — caching responses, invalidating on mutations, showing loading states, and handling errors. Without tag-based invalidation, a TTL-only cache causes stale-self-action bugs in mutation-heavy apps.

**What:** Client-side tagged navigation cache, mutation form actions with auto-invalidation, pending UI with 150ms threshold, and error boundaries — completing the Phase 2 feature set.

**How:**

*Tagged cache:* Replace single `preloadCache` with `Map<string, CacheEntry>` keyed by request path. Each entry stores `{ data, tags, ts, staleTime }`. Reverse tag index (`Map<tag, Set<path>>`) enables O(1) invalidation. LRU via delete+reinsert on hit. Preload hover fetches write to the unified cache with a 5s freshness floor. Config: `Router.start({ cache: { staleTime, gcTime, maxEntries } })`. Public API: `Router.invalidateTags(tags[])`, `Router.invalidate(path?)`.

*Mutation actions:* Delegated `submit` listener intercepts `<form method=post>`. Walks `composedPath()` to find the form and nearest `<webui-route>` component. Calls `static action({ formData, params, signal })` on the component class. After completion, merges action-returned tags with the route's build-time `invalidates` attribute and calls `invalidateTags()`. Dispatches `webui:route:action-complete` event. Supports optimistic state via `result.state`.

*Pending UI:* 150ms `setTimeout` before mounting the pending component in the parent route's outlet area. Skips for keep-alive (instant activation) and cached routes (no fetch). `commitWithData` removes pending elements on arrival.

*Error boundaries:* When `fetchPartial` returns null (HTTP error), finds the error component from the active chain or SSR'd DOM, mounts it in the outlet area with `{ error, status, path }` state. Falls back to `console.warn` if no `error` attribute declared.

## Test Plan

- `pnpm test` in webui-router — 55 unit tests pass (cache hit/miss, tag invalidation, LRU, preload unification)
- Playwright E2E — 4 new tests:
  - Pending UI: slow navigation (500ms delay) shows skeleton after 150ms, then replaces with real content
  - Pending UI: fast navigation skips skeleton entirely
  - Error boundaries: 500 response mounts `error-display` component
  - Error boundaries: successful navigation shows no error component
- E2E fixtures: `loading-skeleton`, `error-display`, `page-slow` (pending attr), `page-failing` (error attr)
- `cargo test --workspace` — 909 Rust tests pass

## Checklist

- [x] Tagged cache with reverse tag index and LRU eviction
- [x] Cache + preload unified (hover preloads use same store)
- [x] `Router.invalidateTags()` and `Router.invalidate()` public API
- [x] Form interception with `composedPath()` for shadow DOM
- [x] Auto-invalidation merges action tags + build-time `invalidates`
- [x] `webui:route:action-complete` event dispatched
- [x] 150ms pending threshold, skip for keep-alive/cached
- [x] Error boundary mounts with `{ error, status, path }` state
- [x] New types exported: `CacheConfig`, `RouteActionContext`, `RouteActionResult`, `ActionCompleteEvent`
- [x] E2E test fixtures and Playwright tests added

> **Stack: 6/6** — depends on #258. Completes the Phase 2 router feature set.

### DIFF
--- a/crates/webui-handler/src/lib.rs
+++ b/crates/webui-handler/src/lib.rs
@@ -392,26 +392,7 @@ impl WebUIHandler {
                 if matched_child.keep_alive {
                     context.writer.write(" keep-alive")?;
                 }
-                if !matched_child.cache_tags.is_empty() {
-                    context.writer.write(" cache-tags=\"")?;
-                    context.writer.write(&matched_child.cache_tags.join(","))?;
-                    context.writer.write("\"")?;
-                }
-                if !matched_child.invalidates.is_empty() {
-                    context.writer.write(" invalidates=\"")?;
-                    context.writer.write(&matched_child.invalidates.join(","))?;
-                    context.writer.write("\"")?;
-                }
-                if !matched_child.pending_component.is_empty() {
-                    context.writer.write(" pending=\"")?;
-                    context.writer.write(&matched_child.pending_component)?;
-                    context.writer.write("\"")?;
-                }
-                if !matched_child.error_component.is_empty() {
-                    context.writer.write(" error=\"")?;
-                    context.writer.write(&matched_child.error_component)?;
-                    context.writer.write("\"")?;
-                }
+                route_renderer::write_route_cache_attrs(context.writer, matched_child)?;
                 context.writer.write(" active>")?;
 
                 context.writer.write("<")?;
@@ -462,26 +443,7 @@ impl WebUIHandler {
                 if child.keep_alive {
                     context.writer.write(" keep-alive")?;
                 }
-                if !child.cache_tags.is_empty() {
-                    context.writer.write(" cache-tags=\"")?;
-                    context.writer.write(&child.cache_tags.join(","))?;
-                    context.writer.write("\"")?;
-                }
-                if !child.invalidates.is_empty() {
-                    context.writer.write(" invalidates=\"")?;
-                    context.writer.write(&child.invalidates.join(","))?;
-                    context.writer.write("\"")?;
-                }
-                if !child.pending_component.is_empty() {
-                    context.writer.write(" pending=\"")?;
-                    context.writer.write(&child.pending_component)?;
-                    context.writer.write("\"")?;
-                }
-                if !child.error_component.is_empty() {
-                    context.writer.write(" error=\"")?;
-                    context.writer.write(&child.error_component)?;
-                    context.writer.write("\"")?;
-                }
+                route_renderer::write_route_cache_attrs(context.writer, child)?;
                 context
                     .writer
                     .write(" style=\"display:none\"></webui-route>")?;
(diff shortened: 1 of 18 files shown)


janechu (commented):
A few issues worth addressing:

---

### 🔴 Bug: `storeCacheEntry()` leaks stale tag-index references on overwrite

When a path is re-fetched and `storeCacheEntry()` is called again for the same `requestPath`, `this.cache.set()` replaces the old `CacheEntry` — but the **old entry's tags are never removed** from `tagIndex`. Over time, `invalidateTags()` will try to evict paths for tags they no longer belong to, and `tagIndex` accumulates stale references.

**Fix:** Call `this.evictCacheEntry(requestPath)` at the top of `storeCacheEntry()` before inserting the new entry. This cleans up old tag-index mappings before establishing new ones.

---

### 🔴 Bug: pending timer isn't cleared when `fetchPartial` throws

The `clearPendingTimer()` on line 1083 is only reached when `fetchPartial` resolves. If it **throws** (network error, DNS failure, offline), the timer is never cleared. A stale 150ms timer can fire after the navigation has already failed, mounting a pending component that is never cleaned up.

Additionally, the error boundary path (lines 1085–1098) is also skipped on throw, so network errors (as opposed to HTTP errors like 500) silently mount a zombie skeleton with no error boundary.

**Fix:** Wrap the fetch + error-boundary block in `try/finally` so `clearPendingTimer()` always runs, and catch network errors to route them into the error boundary:

```ts
try {
  partialData = await this.fetchPartial(requestPath, signal);
} catch (err) {
  if (signal?.aborted) { /* superseded navigation — ignore */ }
  // else: network error — fall through to error boundary below
} finally {
  this.clearPendingTimer();
}
```

---

### 🟡 Action `AbortController` is created but never aborted

In `setupFormInterception()`, a `new AbortController()` is created per form action and `signal` is passed to `ctor.action()`. But nothing ever calls `controller.abort()`. The DESIGN.md explicitly states the signal is "cancelled if the user navigates away during the action."

A late-resolving action can call `setState()` on a component that's been unmounted, invalidate tags for a stale route, and dispatch `webui:route:action-complete` against the wrong page.

**Suggestion:** Track active action controllers and abort them on navigation (`handleNavigation` entry) and `destroy()`.

---

### 🟡 `gcCache()` is defined but never called — `gcTime` is dead config

The `gcCache()` method exists and `gcTime` is accepted in `CacheConfig`, but nothing invokes GC. Entries only leave via LRU eviction or explicit invalidation. Per the DESIGN.md, `gcTime` defaults to 5 minutes — without a timer or sweep, entries can live indefinitely.

**Suggestion:** Either run `gcCache()` periodically (e.g., on each `storeCacheEntry` call, or on a setInterval), or remove `gcTime` from the public config until it's wired up.

---

### 🟡 Minor: `new FormData(form)` drops submit button name/value

Per the HTML spec, `new FormData(form)` without passing `e.submitter` excludes the clicked submit button's `name`/`value`. If a form has multiple submit buttons (e.g., "Save" vs "Delete"), the action handler can't distinguish which was clicked.

**Fix:** `new FormData(form, e.submitter)` (supported in all modern browsers).

---

### 💭 Observation: `PRELOAD_TTL` as a staleTime floor

`lookupCache` uses `Math.max(staleTime, PRELOAD_TTL)` so no entry can expire before 5 seconds. This makes sense for preload hover-to-click, but it means a user-configured `staleTime: 2000` silently becomes 5000ms. May be worth documenting this floor in the `CacheConfig` JSDoc, or only applying the floor for entries stored via preload hover (not via `handleNavigation`).

mohamedmansour:
I am going to squash this with #258

PR #275, opened 2026-04-27
fix(docs): Wrong publish directory

mohamedmansour:
It was not publishing in a sub folder so github pages wasn't working, attempting to fix it


akroshg:
This is fixing this broken page: https://microsoft.github.io/webui/, right?

akroshg (approved):

PR #274, opened 2026-04-26
feat(press): replace VitePress with WebUI Press, a self-hosted SSG

mohamedmansour:
## Description

Introduce `microsoft-webui-press`, a new static-site generator crate that renders the docs.microsoft.github.io/webui site using the WebUI Framework itself. Migrates the existing `/docs` site off VitePress onto Press, and removes the VitePress toolchain entirely. The generator runs at build time to produce static HTML + a binary protocol payload, with WASM-based hydration on the client so the site remains fully static and hostable on GitHub Pages with no backend.

13x faster at SSG Generation than VitePress.

## Why

The WebUI repo was shipping its own documentation on VitePress — a Vue/JS SSG that has nothing to do with WebUI. That undermined the project's core pitch ("language-agnostic SSR without a JS runtime") and meant we had no real-world dogfood of the framework powering a production site.

WebUI Press fixes this by:

- Eating our own dog food: every page on the docs site is rendered by the same parser/handler/protocol pipeline our users are evaluating.
- Demonstrating WASM-based client hydration on a real, GitHub-Pages-hosted static site (no backend, no Node runtime).
- Giving us a reusable "WebUI Framework Powered Static Site Generator" crate that other projects can adopt — not just an internal docs script.

### New crate: `crates/webui-press`

- `microsoft-webui-press` package, `webui-press` binary.
- Pipeline: markdown → frontmatter + syntect highlight → DSD pre-expansion for Web Components in content → parallel page render via rayon → static HTML + protocol payload + search index.
- Default config path: `.webui-press/config.json` (no `--config` flag needed for typical use).
- Custom-page support with per-page `stateFile` whose top-level JSON keys are flattened onto page state, enabling component-driven pages (e.g. the playground's tab list) to SSR from JSON.
- Built-in components: `code-block`, `webui-tabs`/`webui-tab`/ `webui-tab-panel`, `webui-blockquote`.
- Default template (`template/index.html`, `template/docs.css`) with built-in `docs-search` and `docs-theme-toggle` components.

### Site migration: `docs/`

- Removed: `docs/.vitepress/**`, `docs/pnpm-lock.yaml`, npm dev deps.
- Added: `docs/.webui-press/{config.json,theme.css,components,public,state}`.
- Component ports of VitePress Vue components → WebUI components:
  - `CodeComparison.vue`  → `code-comparison`
  - `Playground.vue`      → `docs-playground` (with state-file-driven
                              tabs that SSR from JSON, hydrated to a full
[...]

PR #276, opened 2026-04-28
refactor(dev-server): share rebuild loop between webui-press and webui-cli

mohamedmansour:
Extract a new webui-dev-server crate carrying the watcher, livereload SSE backend, rebuild worker, static-file handler, path helpers, and a TTY-aware RebuildReporter. webui-press and webui-cli now consume the same primitives, so save+refresh behavior, debouncing semantics, and rebuild logging are identical across both servers.

The reporter writes to stderr with a rolling line in interactive terminals and a newline-terminated line under wrappers like concurrently or CI, and reports elapsed time in milliseconds for sub-second rebuilds.


akroshg:
## 🔍 Code Review — PR #276: `refactor(dev-server): share rebuild loop between webui-press and webui-cli`

### What This PR Does

This PR extracts a new **`webui-dev-server`** crate from code previously duplicated (or independently implemented) in `webui-cli serve` and `webui-press serve`. It establishes a shared, composable dev-server toolkit so both tools use identical watcher → rebuild → livereload behavior.

### Architecture

The new crate (`crates/webui-dev-server`) is organized as **six focused modules**, each owning one concern:

```
webui-dev-server/
├── lib.rs          — public re-exports
├── watch.rs        — debounced filesystem watcher (wraps notify-debouncer-mini)
├── worker.rs       — generic rebuild-worker: tick coalescing, rebuild closure, reporter+broadcast
├── livereload.rs   — SSE broadcaster (tokio::broadcast → actix streaming) + client script gen
├── inject.rs       — HTML <script> injection before </body>
├── path.rs         — segment-aware basePath stripping + traversal-safe path resolution
├── serve.rs        — actix static-file handler with basePath routing + livereload injection
└── reporter.rs     — TTY-aware rebuild UX (rolling line on terminals, newline-terminated in CI)
```

**Data flow:**

```
filesystem event → notify-debouncer-mini (50ms coalesce)
  → WatcherHandle.on_event() filters ignored paths
    → TickSender.try_send(()) into sync_channel(8)
      → worker thread: drain ticks → run rebuild closure → report result
        → LiveReload.broadcast_reload() / broadcast_error()
          → SSE stream → EventSource in browser → location.reload()
```

**Consumer integration pattern:**

- **webui-press** (`serve.rs`): composes `spawn_watcher` + `spawn_rebuild_worker` + `serve_static_file`. Rebuilds call `build_docs_with_cache` (reuses syntect highlighter). Serves pre-built static files from `out_dir`.
- **webui-cli** (`commands/serve.rs`): uses `spawn_watcher` + `LiveReload` + `sse_handler` directly. Does NOT use `serve_static_file` — it renders on-the-fly via the WebUI handler. The rebuild worker pattern is inlined since CLI's rebuild updates `Arc<Mutex<SharedState>>` rather than disk files.

**Key design decisions:**

1. **`sync_channel(8)` tick coalescing** — bounded so watcher bursts can't unboundedly queue. Worker drains all before each build. Failed `try_send` is intentional coalescing.
2. **`Arc<Mutex<BuildCache>>`** — only the worker thread mutates it; mutex is uncontended in practice.
3. **TTY detection** — reporter captures `is_terminal()` once at construction, uses `\r\x1b[2K` rolling line on TTY, plain newlines otherwise.
4. **Traversal-safe path resolution** — rejects `..`, `.`, `\`, NUL, consecutive `//`, percent-encoded `..`, and absolute markers per segment.

---

### Review Findings

---

### ISSUE-001: `unsafe` block in `reporter.rs` — `local_hms()` uses `libc::localtime_r`

**Severity**: 🟠 HIGH
**File**: `crates/webui-dev-server/src/reporter.rs` (lines ~147-155)
**Category**: Safety

**Problem**: The workspace-level `Cargo.toml` has `unsafe_code = "deny"` in `[workspace.lints.rust]`, but `reporter.rs` uses an `unsafe` block calling `libc::localtime_r`. This will fail the workspace clippy/lint gate unless the crate opts out.

```rust
unsafe {
    let mut tm: libc::tm = std::mem::zeroed();
    let t: libc::time_t = secs as libc::time_t;
    if !libc::localtime_r(&t, &mut tm).is_null() {
```

The `// SAFETY:` comment is present and correct, but the lint deny will still fire.

**Impact**: `cargo xtask check` will fail on this crate.

**Fix**: Either add `#[allow(unsafe_code)]` on the `local_hms` function with justification, or switch to a safe alternative (e.g., `time` crate's `OffsetDateTime::now_local()`), or add a crate-level lint exception in `Cargo.toml`:
```toml
[lints.rust]
unsafe_code = "allow"  # localtime_r for wall-clock timestamps
```

---

### ISSUE-002: `is_ignored` deduplication uses `Vec::contains` — O(n²) in watcher callback

**Severity**: 🟡 MEDIUM
**File**: `crates/webui-dev-server/src/watch.rs` (lines ~82-85)
**Category**: Performance

**Problem**: Inside the watcher callback (called on every debounce window), deduplication is done via `paths.contains(&e.path)` which is O(n) per event, making the full loop O(n²) for n events in a window.

```rust
if !paths.contains(&e.path) {
    paths.push(e.path);
}
```

**Impact**: Editor save-all operations can produce hundreds of events in one debounce window. With typical 10-50 files this is fine, but a batch format operation on a large docs site could stall the callback.

**Fix**: Use a `HashSet` for dedup:
```rust
let mut seen = std::collections::HashSet::with_capacity(events.len());
let paths: Vec<PathBuf> = events.into_iter()
    .filter(|e| !is_ignored(&e.path, &ignore))
    .filter(|e| seen.insert(e.path.clone()))
    .map(|e| e.path)
    .collect();
```

---

### ISSUE-003: `serve_file_response` converts HTML via `String::from_utf8_lossy` — replaces invalid bytes silently

**Severity**: 🟡 MEDIUM
**File**: `crates/webui-dev-server/src/serve.rs` (lines ~119-122)
**Category**: Correctness

**Problem**: HTML files are read as raw bytes, then converted with `String::from_utf8_lossy(&bytes)` for livereload injection. Invalid UTF-8 bytes are silently replaced with `U+FFFD`, which could corrupt binary content inside HTML (e.g., embedded SVG with binary data, or a BOM-prefixed file).

```rust
let body = if is_html {
    livereload.inject(&String::from_utf8_lossy(&bytes)).into_bytes()
} else {
    bytes
};
```

**Impact**: Rare in practice (HTML files are almost always UTF-8), but violates the "no silent data corruption" principle.

**Fix**: Try `String::from_utf8` first; only inject livereload if the file is valid UTF-8:
```rust
let body = if is_html {
    match String::from_utf8(bytes) {
        Ok(s) => livereload.inject(&s).into_bytes(),
        Err(e) => e.into_bytes(), // serve raw if not valid UTF-8
    }
} else {
    bytes
};
```

---

### ISSUE-004: `resolve_safe_path` doesn't reject Windows drive letters in segments

**Severity**: 🟡 MEDIUM
**File**: `crates/webui-dev-server/src/path.rs` (line ~71)
**Category**: Security

**Problem**: The path resolution rejects `..`, `.`, `\`, NUL, and leading `/`, but doesn't reject Windows drive letters (e.g., `C:` or `C:\`). While `starts_with('/')` catches Unix absolute paths, a segment like `C:` would be pushed onto the PathBuf and could resolve to an absolute path on Windows.

```rust
if s.starts_with('/') {
    return None;
}
buf.push(s);
```

**Impact**: On Windows, `PathBuf::push("C:")` replaces the entire path. A request like `/webui/C:/Windows/System32/config` could escape the output directory.

**Fix**: Add a Windows drive-letter check:
```rust
if s.starts_with('/') {
    return None;
}
// Reject Windows drive letters (e.g., "C:" or "C:\")
if s.len() >= 2 && s.as_bytes()[1] == b':' && s.as_bytes()[0].is_ascii_alphabetic() {
    return None;
}
```

---

### ISSUE-005: `webui-cli serve.rs` — file watcher rebuild doesn't use `spawn_rebuild_worker`

**Severity**: 🟡 MEDIUM
**File**: `crates/webui-cli/src/commands/serve.rs`
**Category**: Design / Inconsistency

**Problem**: The PR description says "webui-press and webui-cli now consume the same primitives, so save+refresh behavior, debouncing semantics, and rebuild logging are identical across both servers." However, webui-cli's `start_file_watcher` does NOT use `spawn_rebuild_worker` — it inlines its own rebuild loop with `std::thread::spawn` + a hand-rolled channel pattern, duplicating the tick-coalescing and error-reporting logic the new crate provides.

**Impact**: The two dev servers will diverge on rebuild UX (no rolling `↻ rebuilt in Xms` line in CLI, different error formatting). The stated goal of the PR — identical behavior — is only partially achieved.

**Fix**: This may be intentional since CLI renders on-the-fly and updates `SharedState` rather than rebuilding to disk. If so, document this explicitly. If not, refactor to use `spawn_rebuild_worker` with a closure that updates `SharedState`.

---

### ISSUE-006: `worker.rs` — `self_tick` clone of sender keeps channel alive after all external senders drop

**Severity**: 🟡 MEDIUM
**File**: `crates/webui-dev-server/src/worker.rs` (line ~56)
**Category**: Correctness

**Problem**: The worker clones the `tx` sender before moving into the thread:
```rust
let self_tick = tx.clone();
thread::spawn(move || {
```
This means even if every external sender is dropped, `self_tick` inside the worker keeps the channel alive. The `rx.recv()` at the top of the loop will never return `Err` while `self_tick` exists — the worker thread becomes a zombie.

**Impact**: When the watcher is dropped (e.g., on Ctrl+C), the worker thread doesn't exit — it blocks forever on `rx.recv()` because `self_tick` holds the channel open.

**Fix**: Use a weak sender pattern, or explicitly signal shutdown:
```rust
// Option A: Drop self_tick when channel has no external senders
// (requires counting senders or using a separate shutdown signal)

// Option B: Use a oneshot channel for shutdown
let (shutdown_tx, shutdown_rx) = std::sync::mpsc::channel::<()>();
```

Or: check `tx.receiver_count()` isn't available on `SyncSender`, so consider using `crossbeam` or a separate `AtomicBool` shutdown flag.

---

### ISSUE-007: No `DESIGN.md` update for the new crate

**Severity**: 🟡 MEDIUM
**File**: `DESIGN.md`
**Category**: Documentation

**Problem**: A new public crate (`webui-dev-server`) was added to the workspace, but `DESIGN.md` — the living specification — was not updated to include it in the architecture diagram or module inventory.

**Impact**: Spec/code divergence. New contributors won't know this crate exists when reading the specification.

**Fix**: Add a `webui-dev-server` entry to the crate inventory in DESIGN.md, describing it as a shared dev-server toolkit consumed by webui-cli and webui-press.

---

### ISSUE-008: `BuildCache` uses `Arc<Mutex<_>>` — comment says "uncontended in practice" but no assertion

**Severity**: 🔵 LOW
**File**: `crates/webui-press/src/serve.rs` (line ~108)
**Category**: Reliability

**Problem**: The cache is wrapped in `Arc<Mutex<_>>` with the comment "the only contender is the worker itself, so the mutex is uncontended in practice." However, if a panic occurs inside the rebuild closure, the mutex becomes poisoned and all subsequent rebuilds fail silently with `"cache mutex poisoned"`.

```rust
let mut guard = cache
    .lock()
    .map_err(|_| "cache mutex poisoned".to_string())?;
```

**Fix**: Use `lock().unwrap_or_else(|e| e.into_inner())` to recover from poisoned mutex (the cache is best-effort state), or switch to `parking_lot::Mutex` which doesn't poison.

---

### Positive Observations ✅

1. **Excellent test coverage** — `watch.rs`, `livereload.rs`, `inject.rs`, `path.rs`, `reporter.rs`, and `serve.rs` all have thorough `#[cfg(test)]` modules with edge cases (traversal, Unicode, TTY state, subscriber lifecycle).
2. **Security-conscious path resolution** — `resolve_safe_path` is well-designed: percent-decode, reject traversal, reject backslashes, reject NUL. The test suite covers `%2e%2e` encoded traversal.
3. **Clean module boundaries** — each module has one concern with a clear doc comment explaining the contract.
4. **Workspace dependency compliance** — all new deps (`notify`, `notify-debouncer-mini`, `percent-encoding`, `futures-util`, `async-stream`) properly use `{ workspace = true }`.
5. **`comrak`, `syntect`, `serde_yaml` moved to workspace deps** — fixes ISSUE-010 from PR #274 review.
6. **`#[must_use]` on appropriate functions** (`normalize_base_path`, `strip_base_path`, `resolve_safe_path`, `LiveReload` getters).

---

## Summary Table

| # | Severity | Category | Title |
|---|----------|----------|-------|
| 1 | 🟠 HIGH | Safety | `unsafe` block in `reporter.rs` will fail workspace `unsafe_code = "deny"` lint |
| 2 | 🟡 MEDIUM | Performance | O(n²) dedup in watcher callback via `Vec::contains` |
| 3 | 🟡 MEDIUM | Correctness | `from_utf8_lossy` silently replaces invalid bytes in HTML |
| 4 | 🟡 MEDIUM | Security | `resolve_safe_path` doesn't reject Windows drive letters |
| 5 | 🟡 MEDIUM | Design | webui-cli doesn't actually use `spawn_rebuild_worker` |
| 6 | 🟡 MEDIUM | Correctness | Worker thread can't exit — `self_tick` keeps channel alive |
| 7 | 🟡 MEDIUM | Documentation | No DESIGN.md update for new crate |
| 8 | 🔵 LOW | Reliability | Poisoned mutex kills all subsequent rebuilds |

## Top 5 Recommendations

1. **ISSUE-001** — Resolve the `unsafe_code = "deny"` conflict before merging — this will fail CI.
2. **ISSUE-004** — Add Windows drive-letter rejection to `resolve_safe_path` — path traversal on Windows is a real security concern for a dev server.
3. **ISSUE-006** — Fix the worker zombie thread — `self_tick` preventing clean shutdown will cause issues in test harnesses and IDE integrations that programmatically stop the server.
4. **ISSUE-005** — Either refactor webui-cli to use `spawn_rebuild_worker`, or update PR description to clarify the architectural split (webui-press uses the full stack; webui-cli uses only watcher + livereload).
5. **ISSUE-007** — Update DESIGN.md with the new crate entry.

akroshg (dismissed):

mohamedmansour:
@akroshg pushed latest commit that resolves the PR.

akroshg (approved):

PR #279, opened 2026-04-29
feat: split rename fast plugin to fast-v2 and add a fast-v3 plugin

janechu:
## Summary

- Keep `fast` as a deprecated compatibility alias and add explicit `fast-v2` for @microsoft/fast-element 2.x marker output.
- Add `fast-v3` as a separate @microsoft/fast-element 3.x plugin path with compact hydration markers.
- Keep the @microsoft/fast-element 2.x and 3.x parser/handler implementations in separate files with no shared plugin helpers, constants, or implementation functions.
- Update the FAST examples, integrations, and docs to use `fast-v3` for @microsoft/fast-element 3.x and refresh the package dependency accordingly.
- Use major-version wording in docs (`@microsoft/fast-element 2.x` / `@microsoft/fast-element 3.x`) instead of specific published versions.

## Validation

- `cargo xtask check`
- `cd docs && pnpm build`


pnpm-workspace.yaml
```diff
@@ -13,8 +13,7 @@ catalog:
   '@codemirror/state': ^6.5.4
   '@codemirror/theme-one-dark': ^6.1.3
   '@codemirror/view': ^6.39.15
-  '@microsoft/fast-element': ^2.10.2
-  '@microsoft/fast-html': ^1.0.0-alpha.39
+  '@microsoft/fast-element': 3.0.0-rc.1
```
mohamedmansour, line 16: We cannot do this, Edge uses fast-element and fast-html, you need to create a new plugin instead of breaking the existing `fast` plugin that @akroshg  is using.

mohamedmansour (changes requested):

pnpm-workspace.yaml
janechu: I'll version it, the previous plugin will be fast-v2, next will be fast-v3

crates/webui-handler/src/plugin/fast.rs
```diff
@@ -26,26 +37,23 @@ use serde_json::Value;
...
-const BINDING_SUFFIX: &str = "$$fe-b-->";
-const SEPARATOR: &str = "$$";
-const REPEAT_START_PREFIX: &str = "<!--fe-repeat$$start$$";
-const REPEAT_END_PREFIX: &str = "<!--fe-repeat$$end$$";
-const REPEAT_SUFFIX: &str = "$$fe-repeat-->";
-const ATTR_SINGLE_PREFIX: &str = " data-fe-b-";
-const ATTR_MULTI_PREFIX: &str = " data-fe-c-";
-
-/// FAST Hydration handler plugin.
+// FAST 2 comment format constants
```
mohamedmansour: I would create a new plugin file, instead of modify the existing deprecated one. Easier to fix bugs, and have isolations.

mohamedmansour (changes requested):

crates/webui-handler/src/plugin/fast.rs
janechu: That is what I asked it to do 😐 argh

mohamedmansour (approved):

PR #286, opened 2026-04-30
fix(framework): trust SSR conditional content during hydration

mohamedmansour:
When hydrating a conditional block, the framework previously required the runtime condition value to be truthy before wiring up SSR content. This caused duplicate DOM when a condition depended on complex properties (like :data) set by the parent AFTER the child hydrates:

1. SSR renders content (condition was true on server)
2. Child hydrates — complex prop not yet set — condition evaluates false
3. SSR DOM nodes left orphaned (not wired, not removed)
4. Parent sets complex prop — condition becomes true — new nodes created
5. Result: duplicate content visible to the user

Fix: during SSR hydration, if a  marker has content after it (the server rendered the block), hydrate that content regardless of the current condition value. The server was authoritative; the data will arrive from the parent and the condition will reconcile correctly.

Added helper to detect SSR content between conditional markers.

### DIFF
--- a/packages/webui-framework/src/element.ts
+++ b/packages/webui-framework/src/element.ts
@@ -733,8 +733,18 @@ export class WebUIElement extends HTMLElement {
         }
         if (marker) lastCondMarker = marker;
 
-        if (shown && blockMeta && marker) {
-          condInstance = this.$hydrateCondContent(condAnchor, blockMeta, scope);
+        // SSR hydration: if the marker exists, the server rendered this
+        // conditional — hydrate its content regardless of the current
+        // condition value.  Complex properties from parent bindings may
+        // not have arrived yet (the parent hydrates after children), so
+        // the condition could evaluate to false even though the server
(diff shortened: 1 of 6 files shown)


packages/webui-framework/src/element.ts
```diff
@@ -989,6 +1002,24 @@ export class WebUIElement extends HTMLElement {
...
+  /**
+   * Check whether there is non-marker content between a conditional
+   * start anchor and its closing marker.  Used during SSR hydration to
+   * detect server-rendered conditional content even when the runtime
+   * condition value has not been set yet (e.g. complex property from a
+   * parent repeat binding that hydrates after its children).
+   */
+  private $hasContentAfterMarker(anchor: Comment, endData: string): boolean {
+    let sibling = anchor.nextSibling;
+    while (sibling) {
```
akroshg, line 1014: WE would loop only once, why while loop?

akroshg (approved):

PR #288, opened 2026-05-01
feat: preserve shadowroot* attrs on component template wrapper

janechu:
## Summary

When `DomStrategy::Shadow` is active, the WebUI parser now preserves any `shadowroot`-prefixed attributes the user wrote on their wrapping `<template>` and applies these rules:

- **Mode** — `shadowrootmode` value (or, if absent, a bare legacy `shadowroot=` attribute) is captured. Both attributes are then **always** emitted on the wrapper as `shadowrootmode="<mode>" shadowroot="<mode>"` (the legacy `shadowroot` attribute is emitted as a paired alias for older user agents / tooling that look for either form). Defaults to `"open"` when neither is supplied.
- **Other shadow-root attrs** — `shadowrootclonable`, `shadowrootdelegatesfocus`, `shadowrootserializable`, etc. are preserved verbatim on the emitted `<template>`.
- **`shadowrootadoptedstylesheets`** — when the user supplies a value AND the CSS Module strategy would also generate one, the values are **merged** into a single space-separated list (user-supplied specifiers first in their original order; parser-generated specifiers appended only if not already present, deduplicated).
- **Runtime / non-shadowroot user attrs on `<template>`** — `@event`, `:prop`, `?bool`, and any other non-`shadowroot*` attribute on the user's `<template>` is stripped from the emitted wrapper (existing behaviour).

Under the **FAST v2 / v3 parser plugins** those `shadowroot*` attributes are placed on the outer **`<f-template>`** element rather than the inner `<template>`. `<f-template>` is the framework-owned transport element that the FAST runtime consumes; the inner `<template>` always has all `shadowroot*` attributes stripped so the browser does not auto-activate it as a declarative shadow root inside `<f-template>`. The same `shadowrootadoptedstylesheets` merge rule applies on `<f-template>`.

## Bug fix included

Also fixes a pre-existing bug in `generate_f_template` (fast-v2 / fast-v3) that used `find('>')` to locate the end of the outer `<template>` opening tag — this broke on attribute values containing `>` (e.g. `data-note="a > b"`). The new implementation uses the existing `find_tag_close` helper which respects quoted attribute values, with a regression test in both plugins.

## Backward compatibility

Light-DOM mode is unaffected: the `<template>` wrapper continues to be stripped entirely. The default shadow-DOM behaviour (no user `shadowroot*` attrs supplied) emits `<template shadowrootmode="open" shadowroot="open">`, which is backward-compatible with existing apps that rely on `shadowrootmode="open"`.

## Tests

- 8 new tests in the parser core (`crates/webui-parser/src/lib.rs`) covering: default paired emission, user-supplied `shadowrootmode="closed"`, bare legacy `shadowroot=` promoted to mode, `shadowrootclonable`/`shadowrootdelegatesfocus` preservation, adopted-stylesheets merge with user value, dedup against module specifier, runtime attrs stripped from wrapper, and the light-DOM regression.
- 7 new tests in each of `fast_v2.rs` and `fast_v3.rs` (14 total) covering: `shadowrootmode` moved to `<f-template>`, user closed mode preserved, other shadow-root attrs on `<f-template>`, no `shadowroot*` on `<f-template>` when user did not supply any, module CSS strategy puts `shadowrootadoptedstylesheets` on `<f-template>`, merged with user value, and the `>` in quoted attr regression test. Plus updated existing tests to assert the new placement and a coverage test for the renamed `strip_shadowroot_attrs` helper.
- All 386 parser tests pass.
- Full `cargo xtask check` passes (license-headers, fmt, clippy, deny, test, build, build (wasm), build (examples), bench, docs).

## Documentation

- `DESIGN.md` — new "Shadow DOM `<template>` wrapper" subsection documenting the preservation, paired-alias, merge, and FAST-plugin placement rules.
- `docs/guide/concepts/components/index.md` — updated component-author guidance.
- `docs/guide/cli/index.md` — updated DOM strategy table.
- `docs/ai.md` — updated AI-targeted guidance.
- `packages/webui-framework/README.md` and `RENDERING.md` — example output reflects paired `shadowrootmode`/`shadowroot`.
- Doc-comments on `process_component_template`, `generate_f_template`, `convert_btr_to_fast`, and `strip_shadowroot_attrs` describe the new behaviour.

### DIFF
--- a/crates/webui-parser/src/lib.rs
+++ b/crates/webui-parser/src/lib.rs
@@ -1808,14 +1808,24 @@ impl HtmlParser {
 
     /// Process component template HTML.
     ///
-    /// - **Shadow DOM** (`DomStrategy::Shadow`): wraps content in
-    ///   `<template shadowrootmode="open">`, preserves `:host` CSS,
-    ///   optionally adds `shadowrootadoptedstylesheets`.
+    /// - **Shadow DOM** (`DomStrategy::Shadow`): wraps content in a
+    ///   `<template>` with declarative-shadow-DOM attributes. Any
+    ///   `shadowroot`-prefixed attributes on the user-supplied wrapping
+    ///   `<template>` (`shadowrootmode`, `shadowrootclonable`,
+    ///   `shadowrootdelegatesfocus`, `shadowrootserializable`,
+    ///   `shadowrootadoptedstylesheets`, or a bare legacy `shadowroot`)
+    ///   are preserved on the emitted wrapper. Both `shadowrootmode` and
+    ///   the legacy `shadowroot` attribute are always emitted as a matching
+    ///   pair (defaulting to `"open"` when the user did not supply a mode).
+    ///   When the parser also generates an adopted-stylesheet specifier
+    ///   (CSS Module strategy), it is merged with the user's
+    ///   `shadowrootadoptedstylesheets` value as a space-separated,
+    ///   de-duplicated list. `:host` CSS is preserved.
     /// - **Light DOM** (`DomStrategy::Light`): strips any existing shadow DOM
     ///   wrapper, outputs plain HTML.
     ///
-    /// In both modes, runtime-only attributes (`@`, `:`, `?`) are stripped
-    /// from the opening `<template>` tag.
+    /// In both modes, runtime-only attributes (`@`, `:`, `?`) and any other
+    /// non-`shadowroot*` attributes on the user's `<template>` are stripped.
     fn process_component_template(
         &mut self,
         html: &str,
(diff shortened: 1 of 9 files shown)


crates/webui-parser/src/lib.rs
```diff
@@ -1952,6 +2005,75 @@ impl HtmlParser {
     }
 }
 
+/// Captured attribute state from a user-supplied `<template>` opening tag.
+///
+/// Used by [`HtmlParser::extract_template_attrs`] to surface
+/// `shadowroot`-prefixed attributes so they can be re-emitted on the
+/// declarative-shadow-DOM wrapper.
+#[derive(Default, Debug)]
+struct ExtractedTemplateAttrs {
```
mohamedmansour, line 2014: I don't want us to capture each individual user supplied attributes, we should preserve ALL the attributes if the webui component supplied them.

mohamedmansour (changes requested):

mohamedmansour:
This is already implemented where we preserve all template attributes when provided

PR #290, opened 2026-05-05
Add entry_point build option for script tag injection

### LINKED ISSUE #289: Add entry_point build option for script tag injection
## Problem

The Rust WebUI build path does not embed the entry-point script tag in the protocol. Consumers (C++ host, JS dev-server) must post-process rendered HTML to manually inject the tag before \</body>\.

## Solution

Add an optional \ntry_point\ field to \BuildOptions\. When provided, the parser injects \<script type=\"module\" src=\"./NAME.js\">\</script>\ as a raw fragment immediately before the \ody_end\ signal. When \None\, behavior is unchanged (backward compatible).

## Changes

- \webui-parser\: Add \ntry_point\ field + inject raw fragment before \ody_end\
- \webui\ (build API): Add \ntry_point: Option<String>\ to \BuildOptions\, wire to parser
- \webui-node\: Expose \ntryPoint\ in \JsBuildOptions\
- DESIGN.md and docs updated

akroshg:
## Summary

Add an optional \ntry_point\ field to \BuildOptions\ that injects a \<script type=\"module\" src=\"./NAME.js\">\</script>\ raw fragment into the protocol immediately before the \ody_end\ signal. This eliminates the need for consumers to post-process rendered HTML to inject entry scripts.

Closes #289

## Changes

| Layer | Change |
|-------|--------|
| \webui-parser\ | Add \ntry_point\ field + \set_entry_point()\ setter + inject raw fragment before \ody_end\ |
| \webui\ (build API) | Add \ntry_point: Option<String>\ to \BuildOptions\, wire to parser |
| \webui-node\ | Expose \ntryPoint\ in \JsBuildOptions\ |
| DESIGN.md | Add \ntry_point\ to \HtmlParser\ struct listing |
| docs/ | Add \ntry_point\/\ntryPoint\ to Rust and Node integration docs |

## Testing

- 3 unit tests in \webui-parser\ (injection, no-injection, hyphenated name)
- 2 integration tests in \webui\ crate (BuildOptions with/without entry_point)
- All existing tests pass (pre-existing \webui-discovery\ failure is unrelated)

## Backward Compatibility

\ntry_point: None\ (default) = no script tag injected = current behavior.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -857,6 +857,7 @@ pub struct HtmlParser {
     condition_parser: ConditionParser,
     handlebars_parser: HandlebarsParser,
     css_strategy: CssStrategy,
+    entry_point: Option<String>,
     // Other fields...
 }
 ```
--- a/crates/webui-cli/src/commands/common.rs
+++ b/crates/webui-cli/src/commands/common.rs
@@ -45,6 +45,7 @@ impl AppArgs {
             dom: self.dom,
             plugin: self.plugin,
             components: self.components.clone(),
+            entry_point: None,
         }
     }
 }
--- a/crates/webui-node/src/lib.rs
+++ b/crates/webui-node/src/lib.rs
@@ -79,6 +79,8 @@ pub struct JsBuildOptions {
     pub plugin: Option<String>,
     /// Additional component sources (npm packages or local paths).
     pub components: Option<Vec<String>>,
+    /// JS entry-point name (without `.js`); injects script tag into protocol.
+    pub entry_point: Option<String>,
 }
 
 /// Build a WebUI application from an app directory.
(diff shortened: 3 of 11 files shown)


mohamedmansour (commented):
This doesn't look like the right api given we have "entry". Given we already emit "body_end" signal, why couldn't we just add it to the state? Or why not just add the <script> tag directly in the entry html file since this is static.

mohamedmansour (changes requested):
This doesn't look like the right api given we have "entry". Given we already emit "body_end" signal, why couldn't we just add it to the state? Or why not just add the <script> tag directly in the entry html file since this is static.

akroshg:
Partly for the reason to emit the script tag at build time instead of runtime (state is runtime).
In the edge side we don't put the script tags in the index.html, we do this processing at generator.js prior to the emitting body_end.
However the cost is doing at runtime is very minimal. So let me try that approach and report back.

akroshg:
Closing  the entry script injection is handled via the existing \ody_end\ state mechanism on the Edge side instead. No WebUI framework changes needed.

PR #294, opened 2026-05-15
feat: streaming SSR pipeline

mohamedmansour:
SSR pipeline with bounded-channel streaming, lock-free chunk pool, structural per-render HTML injection, and a zero-allocation hot path. Replaces the legacy buffer-then-byte-scan-and-concat pipeline.

## Headline numbers

Per-render, contact-book @ 1000 contacts, allocator-exact + getrusage. Origin/main `string+postinject` → this PR `streaming+inject POOLED`:

| metric | origin/main | this PR | **Δ** |
|---|---|---|---|
| allocations | 526 | 520 | −1.1 % |
| **bytes/render** | 75.0 KiB | 30.3 KiB | **−59.6 %** |
| **user CPU µs** | 33.6 | 22.5 | **−33 %** |

Browser-perceived metrics (Playwright, real Chromium, 250 ms render):

| metric | buffered | streaming | **Δ** |
|---|---|---|---|
| **TTFB** | 265 ms | 0.4 ms | **663× faster** |
| **FCP / LCP** | 284 ms | 56 ms | **5.1× faster** |

## Three commits

| # | Commit | What |
|---|---|---|
| 1 | [`86ca1cbd`](../commit/86ca1cbd) | Bench harness (origin/main-compatible: `string`, `string+postinject`). Cherry-pickable onto origin/main for a baseline. |
| 2 | [`7767be48`](../commit/7767be48) | `StreamingWriter` + `ChunkPool` primitive + 3 new bench layers (criterion writer-paths, e2e-ttfb actix, Playwright browser). |
| 3 | [`6e3af609`](../commit/6e3af609) | `RenderOptions::with_head_inject` / `with_body_inject`, dedup DoS guard, 6 hot-path allocation cuts, CLI / commerce wiring. |

## What's in each commit

**Streaming primitive (commit 2)** — bounded tokio mpsc (default 4 chunks ≈ 16 KiB → backpressure), configurable `with_flush_timeout` (slow-loris DoS bound), lock-free `ChunkPool` via `crossbeam_queue::ArrayQueue` for zero per-flush allocation in steady state, typed `ClientDisconnected` / `StreamTimeout` errors.

**Structural HTML injection (commit 3)** — `with_head_inject` / `with_body_inject` emit at parser-synthesized `head_end` / `body_end` structural boundaries. No byte scanner — cannot mis-fire on `</head>` / `</body>` literals in HTML comments, `<iframe srcdoc>`, or inline `<script>`. Dedupe guard prevents inject amplification (a 1 MiB inject × N duplicate signals would otherwise produce N MiB of output).

**Hot-path perf (commit 3)** — `request_path` / `entry_id` / `nonce` / `route_base` switched from `String` to `&'a str` / `Cow<'a, str>` (4 fewer allocations per render); `<for>` loop variable name inserted once + `get_mut`-swapped per iteration (2·(N−1) allocations saved on N-iter loops); per-render `component_index_cache` eliminates the second `build_component_index` walk at `body_end`.

**Security guards (commit 3)** — empty-string normalization at handler init (defends against `RenderOptions { nonce: Some(""), .. }` field-bypass which would emit `<script nonce="">`, a hard CSP failure); explicit XSS warning on inject builder doc comments; `encode_safe` re-exported from `webui_handler` for callers that need pre-escaping.

## Tests

299 total (15 streaming + 284 handler), including: 16-thread concurrent-render stress, 1 MiB inject roundtrip, cross-thread `Bytes` drop, marker-spoof robustness (`<!-- </body> -->` literals), dedupe guard, field-bypass CSP, nested `<for>` reusing same variable name, runtime-free flush timeout positive test, `end()` surfaces first-flush error. `cargo xtask check` ✅ green (1344 s).

## Reproduce

```bash
# Resource bench: per-render allocations / bytes / CPU
git checkout 86ca1cbd && cargo xtask bench streaming-resource --save-baseline c1
git checkout 7767be48 && cargo xtask bench streaming-resource --baseline c1 --save-baseline c2
git checkout 6e3af609 && cargo xtask bench streaming-resource --baseline c2

# Browser metrics in real Chromium
cd examples/integration/streaming-browser-bench && pnpm test
```

## Docs

- `BENCHMARKS.md` — bench layer reference + before/after workflow
- `DESIGN.md` "Streaming Response Writers" — primitive + injection API + safety contracts
- `crates/webui/benches/README.md`, `examples/integration/streaming-browser-bench/README.md`

## Production wiring

- `crates/webui-cli/src/commands/serve.rs` — dev server uses `StreamingWriter::new_pooled` (256-slot pool ≈ 1.25 MiB peak), 30 s flush deadline, livereload script as `Arc<str>` via `with_body_inject`.
- `examples/app/commerce/server` — same pattern; per-page image preload `<link>` via `with_head_inject`.

### DIFF
--- a/crates/webui-handler/src/lib.rs
+++ b/crates/webui-handler/src/lib.rs
@@ -55,6 +55,24 @@ pub enum HandlerError {
 
     #[error("Plugin data error: {0}")]
     PluginData(String),
+
+    /// The HTTP client disconnected before the render completed.
+    ///
+    /// Streaming `ResponseWriter` implementations return this from
+    /// `write()` once their channel/socket is closed, so the handler
+    /// can abort the render rather than do CPU work that has nowhere
+    /// to go. Allocation-free (the variant carries no payload).
+    #[error("client disconnected")]
+    ClientDisconnected,
+
+    /// The streaming writer's flush exceeded its configured deadline.
+    ///
+    /// Indicates a slow/unresponsive consumer (slow-loris client,
+    /// stuck proxy, etc.). The render thread is freed; downstream
+    /// telemetry should distinguish this from `ClientDisconnected`
+    /// so ops can alert on slow-client attacks.
+    #[error("streaming flush timed out")]
+    StreamTimeout,
 }
 
 pub type Result<T> = std::result::Result<T, HandlerError>;
--- a/crates/webui-handler/src/lib.rs
+++ b/crates/webui-handler/src/lib.rs
@@ -12,6 +12,17 @@ pub mod route_handler;
 pub mod route_matcher;
 pub(crate) mod route_renderer;
 
+/// Minimal HTML escaper for the 6 XSS-critical characters
+/// (`& < > " ' /`). Returns `Cow::Borrowed` when no escaping is
+/// needed (zero allocation on the happy path), `Cow::Owned` when
+/// any character had to be replaced.
+///
+/// Re-exported here so external callers of `RenderOptions::with_head_inject`
+/// / `with_body_inject` can pre-escape untrusted content with the
+/// same escaper the handler uses internally for SSR text content,
+/// without having to pull in a separate HTML-escape crate.
+pub use html_encode::encode_safe;
+
 use plugin::HandlerPlugin;
 use route_matcher::CompiledRouteCache;
 use serde::Serialize;
@@ -99,6 +110,19 @@ pub struct RenderOptions<'a> {
     /// When set, all inline scripts include `nonce="VALUE"` and a
     /// `<meta name="webui-nonce">` tag is emitted for the client router.
     pub nonce: Option<&'a str>,
+    /// Optional HTML to emit immediately before the document's
+    /// `</head>` close. Used for per-request `<link rel="preload">`
+    /// hints, CSP `<meta>` tags beyond the built-in nonce, etc.
+    /// Inserted at the structural `head_end` boundary identified by
+    /// the parser — never matched against a byte pattern, so cannot
+    /// be tricked by `</head>` literals appearing in HTML comments,
+    /// `srcdoc` attributes, or inline scripts.
+    pub head_inject: Option<&'a str>,
+    /// Optional HTML to emit immediately before the document's
+    /// `</body>` close. Used for dev livereload `<script>`, analytics
+    /// snippets, OpenTelemetry trace IDs, etc.
+    /// Same structural-boundary guarantee as [`head_inject`](Self::head_inject).
+    pub body_inject: Option<&'a str>,
 }
 
 impl<'a> RenderOptions<'a> {
@@ -109,13 +133,52 @@ impl<'a> RenderOptions<'a> {
             entry_id,
             request_path,
             nonce: None,
+            head_inject: None,
+            body_inject: None,
         }
     }
 
-    /// Set the CSP nonce for inline scripts.
+    /// Set the CSP nonce for inline scripts. Pass an empty string to
+    /// disable (`None` semantics) — empty `<meta name="webui-nonce"
+    /// content="">` would be browser-ignored noise.
     #[must_use]
     pub fn with_nonce(mut self, nonce: &'a str) -> Self {
-        self.nonce = Some(nonce);
+        self.nonce = if nonce.is_empty() { None } else { Some(nonce) };
+        self
+    }
+
+    /// Set HTML to emit immediately before `</head>`.
+    /// Pass an empty string to disable (`None` semantics).
+    ///
+    /// # Safety (XSS warning)
+    ///
+    /// The provided HTML is written verbatim — **no HTML escaping is
+    /// performed**. Callers MUST ensure the content is fully trusted
+    /// (typically a `&'static str` or build-time-derived bytes such as
+    /// dev livereload script, image preload `<link>` tags, or A/B test
+    /// markers). Passing user-controlled or attacker-influenced content
+    /// here is a direct cross-site scripting vulnerability. If your
+    /// caller path may include untrusted data, escape with the host's
+    /// HTML escaper (e.g. [`webui_handler::encode_safe`](crate::encode_safe))
+    /// **before** calling this builder.
+    #[must_use]
+    pub fn with_head_inject(mut self, html: &'a str) -> Self {
+        self.head_inject = if html.is_empty() { None } else { Some(html) };
+        self
+    }
+
+    /// Set HTML to emit immediately before `</body>`.
+    /// Pass an empty string to disable (`None` semantics).
+    ///
+    /// # Safety (XSS warning)
+    ///
+    /// Same contract as [`with_head_inject`](Self::with_head_inject):
+    /// the HTML is written verbatim with **no escaping**, so callers
+    /// MUST ensure the content is fully trusted. Untrusted content is
+    /// a direct XSS vector.
+    #[must_use]
+    pub fn with_body_inject(mut self, html: &'a str) -> Self {
+        self.body_inject = if html.is_empty() { None } else { Some(html) };
         self
     }
 }
@@ -136,11 +199,14 @@ struct WebUIProcessContext<'a> {
     local_vars: HashMap<String, Value>,
     /// Accumulates component attribute values between attrStart and the component fragment.
     component_attrs: HashMap<String, Value>,
-    /// URL path for server-side route matching.
-    request_path: String,
+    /// URL path for server-side route matching. Borrowed from
+    /// `RenderOptions<'a>::request_path` — zero-copy.
+    request_path: &'a str,
     /// Base path for resolving relative route paths (`./`).
     /// Updated as the handler descends into nested matched routes.
-    route_base: String,
+    /// `Cow` keeps the initial `"/"` literal zero-copy; nested-route
+    /// descent owns the recomputed path.
+    route_base: Cow<'a, str>,
     /// Component names visited during rendering (for selective f-template emission
     /// and CSS module dedup — only the first render of each component emits
     /// its `<style type="module">` tag).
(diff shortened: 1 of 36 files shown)


crates/webui/src/streaming.rs
```diff
@@ -0,0 +1,820 @@
...
+        return match timed {
+            Ok(Ok(())) => SendOutcome::Ok,
+            Ok(Err(_)) => SendOutcome::Disconnected,
+            Err(_) => SendOutcome::TimedOut,
+        };
+    }
+    // No runtime: the documented usage requires a runtime when
+    // `with_flush_timeout` is set, so this branch only triggers in
+    // misuse or tests. Fall back to untimed blocking_send.
+    debug_assert!(
```
akroshg: **Slow-loris guard silently disables itself in release when no tokio runtime is in TLS.**

`with_flush_timeout(30s)` is the framework's only slow-loris DoS guard (per `serve.rs` and DESIGN.md: *"the chosen timeout × concurrent-render-limit is the upper bound on resources an attacker can pin"*). When a host wires the writer from anything that isn't a tokio `spawn_blocking` worker (raw `std::thread::spawn`, a Tower service, an FFI bridge, a custom worker pool), `Handle::try_current()` at line 535 returns `Err`. In release the `debug_assert!` here is a no-op, control falls through to `tx.blocking_send` on line 551, and the configured deadline silently disappears — a slow client now pins the producer thread forever, with no log/metric/error. With `panic = "abort"` in the release profile the assert can't help us in production either way.

There's a second crash mode hiding here too: if this code path is reached from a tokio worker that is *not* `spawn_blocking`, `tx.blocking_send` itself panics with *"Cannot block the current thread from within a runtime"*, and `panic = "abort"` then kills the entire process. So the same misconfiguration is either silent infinite block or full process abort depending on host shape.

**Suggestion:** fail loudly instead. Either return `SendOutcome::TimedOut` after a real `std::thread::sleep(deadline)` + `try_send`, or validate runtime presence at builder time and have `with_flush_timeout` return a `Result`. At minimum a `log::warn!("with_flush_timeout configured but no tokio runtime in TLS — slow-loris bound disabled")` would make this visible to operators in production.

Also worth adding: there is no positive test for this path. A `#[tokio::test(flavor = "multi_thread")]` that fills a 1-slot channel without a consumer and asserts `Err(HandlerError::StreamTimeout)` would pin the happy path of this guard, which is currently entirely unverified.

akroshg (commented):
Inline notes on a few things worth a look. The architecture (structural-signal injection, lock-free `ChunkPool`, `Bytes::from_owner` ownership, bounded-channel back-pressure) is the right design and the perf claims hold up — these are 4 edge-case bugs worth fixing on top.

crates/webui/src/streaming.rs
```diff
@@ -0,0 +1,820 @@
...
+        }
+        Ok(())
+    }
+
+    fn end(&mut self) -> Result<()> {
+        // On end, attempt a final flush but never error out: the caller
+        // is finishing the response, and a terminated channel here
+        // means the client gave up.
+        if self.terminated.is_none() {
+            let _ = self.flush_buf();
```
akroshg: **`end()` swallows the final-flush error, silently truncating responses.**

Trace the normal finish path:

1. Render writes 3 KB (under `chunk_target`, no flush yet).
2. `end()` runs, `terminated` is `None`, `flush_buf()` tries to send the trailing 3 KB, hits a `ClientDisconnected`, sets `terminated`, returns `Err`.
3. The `let _ = …` here drops the `Err` and `end()` returns `Ok(())`.

The caller observes a successful end-of-render. Access logs / telemetry can't distinguish "fully delivered" from "client cancelled at the very last chunk" — a 200-OK entry gets recorded for what was actually a truncated response, and there's no signal a `render_errors_total` counter can react to.

This is exactly the failure mode `HandlerError::ClientDisconnected` was introduced for in this PR — the same logic that motivates surfacing it from `write()` ("abort the render rather than do CPU work that has nowhere to go") applies to the final flush. Swallowing in `Drop` is fine; `end()` is a deliberate, fallible API call and its error type exists for a reason.

**Suggestion:**
```rust
fn end(&mut self) -> Result<()> {
    if self.terminated.is_some() { return Ok(()); }
    self.flush_buf()
}
```

crates/webui-handler/src/lib.rs
```diff
@@ -81,6 +99,19 @@ pub struct RenderOptions<'a> {
     /// When set, all inline scripts include `nonce="VALUE"` and a
     /// `<meta name="webui-nonce">` tag is emitted for the client router.
     pub nonce: Option<&'a str>,
```
akroshg, line 112: **Public field bypasses the `"" → None` normalisation in `with_nonce`, which can cause a complete inline-script outage under strict CSP.**

`with_nonce("")` (line 134-137) correctly normalises to `None`, and there's a regression test for it. But because `nonce` is still a `pub` field, a caller can write

```rust
RenderOptions { nonce: Some(""), ..RenderOptions::new(e, p) }
```

…or use the field-update syntax against an existing struct, and the handler will dutifully emit `<meta name="webui-nonce" content="">` plus `nonce=""` on every inline `<script>` / `<style type="module">`.

Under a strict `Content-Security-Policy: script-src 'nonce-…'` header, `<script nonce="">` is a **hard CSP failure** — every inline script is blocked even though the header was set. So the "browser-ignored noise" rationale in the commit message understates the consequence on this code path: it turns into a complete inline-script-execution outage. (For `head_inject` / `body_inject` on lines 109 / 114 the consequence is much milder — empty inserts are no-ops — but normalising them uniformly keeps the contract sane.)

**Suggestion:** either make the fields `pub(crate)` and force the builder, or normalise at handler initialisation:

```rust
nonce: options.nonce.filter(|n| !n.is_empty()),
head_inject: options.head_inject.filter(|s| !s.is_empty()),
body_inject: options.body_inject.filter(|s| !s.is_empty()),
```

crates/webui-handler/src/lib.rs
```diff
@@ -91,13 +122,52 @@ impl<'a> RenderOptions<'a> {
...
+    /// # Safety (XSS warning)
+    ///
+    /// The provided HTML is written verbatim — **no HTML escaping is
+    /// performed**. Callers MUST ensure the content is fully trusted
+    /// (typically a `&'static str` or build-time-derived bytes such as
+    /// dev livereload script, image preload `<link>` tags, or A/B test
+    /// markers). Passing user-controlled or attacker-influenced content
+    /// here is a direct cross-site scripting vulnerability. If your
+    /// caller path may include untrusted data, escape with the host's
+    /// HTML escaper (e.g. `webui_handler::html_encode::encode_safe`)
```
akroshg: **The recommended escaper isn't reachable from outside the crate — this doc points callers at a `pub(crate)` function.**

The doc here recommends `webui_handler::html_encode::encode_safe` for escaping untrusted content before passing it to `with_head_inject` / `with_body_inject`. But at `crates/webui-handler/src/lib.rs:9`:

```rust
pub(crate) mod html_encode;
```

External callers can't import `webui_handler::html_encode::encode_safe` — the documented escape route doesn't compile. The DESIGN.md "Per-Render HTML Injection" section mirrors the same broken reference. Most host authors who hit this will silently roll their own (or skip escaping entirely), which defeats the whole point of the safety note.

**Suggestion:** one of

- `pub use html_encode::encode_safe;` at the crate root (the function is single-purpose and well-defined),
- promote to `pub mod html_encode;`, or
- change the doc to point at a published external escaper (`htmlescape`, `v_htmlescape`, etc.).

Same edit needed in DESIGN.md's "Per-Render HTML Injection" section.

akroshg (dismissed):

mohamedmansour:
Thanks @akroshg — all four findings were valid, fixed in `bdd5fee7`. Also did an adversarial re-audit on top to hunt for whatever both reviews missed; results below.

## ✅ Your 4 findings — all fixed

1. **`with_flush_timeout` silent-disable + abort risk** (`crates/webui/src/streaming.rs:545-606`). Replaced the `debug_assert!(false)` (no-op in release; never could have helped under `panic = "abort"`) with a runtime-free poll loop (`try_send` + `thread::sleep`) that actually enforces the deadline when no tokio runtime is in TLS. A `warn-once` `log::warn!` makes the misconfiguration visible to ops. New positive test `streaming_writer_flush_timeout_fires_without_runtime` verifies `Err(StreamTimeout)` fires within ±1500ms of the configured 150ms deadline and short-circuits subsequent writes.

2. **`end()` swallowing the final-flush error**. `end()` now surfaces `Err(ClientDisconnected)` / `Err(StreamTimeout)` on first-disconnect (returns `Ok` if already terminated, since `write()` already surfaced the error then). New test `streaming_writer_end_surfaces_first_flush_error` pins the contract. Updated production callers (`webui-cli/src/commands/serve.rs`, `examples/app/commerce/server/src/server.rs`) to `log::debug!` truncated-stream errors so they're visible in telemetry without spamming production logs.

3. **`pub` field bypass of `with_nonce("")` normalisation → CSP outage**. Added defensive `.filter(|s| !s.is_empty())` on all three injection points (`nonce`, `head_inject`, `body_inject`) at handler init in both `handle()` and `render()`. Field-update syntax can no longer trigger a hard-CSP `<script nonce="">` failure. New test `empty_field_bypass_is_normalised_at_handler_init` constructs `RenderOptions { nonce: Some(""), .. }` directly (bypassing the builder) and asserts the empty marker never reaches the output.

4. **Broken doc reference to `pub(crate)` symbol**. Added `pub use html_encode::encode_safe;` at the crate root with a doc comment explaining the re-export's purpose. Updated the inline doc and the DESIGN.md section to point at the now-reachable `webui_handler::encode_safe`. Hosts can pre-escape untrusted content with the same escaper the handler uses internally.

## Adversarial re-audit findings

Ran a fresh review specifically looking for patterns matching your four (release-vs-debug gaps, swallowed `let _ = result;`, `pub`-field bypasses, broken doc paths) plus panic-safety / Send+Sync / nested loop / Playwright concerns.

**Bug-6 from the re-audit (nested `<for>` loops reusing the same variable name)** — investigated and **proved invalid**. The audit claimed the outer's `get_mut` held a reference past the recursive `process_fragment_id`, clobbering the inner's restore. Walking through the code: `get_mut` is called fresh each iteration (line 966) and immediately deref-written, so no reference persists across the recursive call. The inner's `remove` + `pre-insert` + `restore` correctly preserves the outer's value for any fragments emitted between the inner loop and the outer's next iteration. Added regression test `nested_for_loops_reusing_same_variable_name_dont_corrupt_scope` that exercises `[(O=A[I=X][I=Y],O=A)(O=B[I=X][I=Y],O=B)]` — proves the outer `item` is bound correctly before, between, and after the inner loop.

**Bug-8 from the re-audit (cumulative `addInitScript` calls in Playwright bench)** — valid 🔵 perf/test bug. The bench calls `measure()` 64 times (8 iters × 4 scenarios × 2 paths), and the original LCP fix put `await page.addInitScript(...)` inside `measure`. Playwright accumulates init scripts per browser context, so by iteration 64, 64 copies of the LCP observer registered on every navigation. Fixed by extracting `ensureLcpObserverInstalled` with a `__lcpObserverInstalled` flag on the context, called once via `addInitScript` on the shared context.

**Bug-9 from the re-audit (`waitForTimeout(300)` for LCP settle is arbitrary)** — valid 🔵 test flakiness. The 300ms wait wasn't tied to any LCP-stable signal; on a contended CI runner with a `delay_us=500` (~250ms render) scenario, the LCP candidate could keep updating past 300ms and the test would read an incomplete `__lcpEntries` array. Replaced with `waitForLcpStable`: polls `__lcpEntries.length` every 50ms, returns once it's been stable for 200ms, capped at 2000ms so the test cannot hang on adversarial pages.

## Net change

- **9 total findings** across both reviews (4 yours + 5 from re-audit), of which **8 valid** (4 yours all confirmed; 1 from re-audit confirmed invalid via regression test; 4 from re-audit confirmed valid and fixed).
- **3 new regression tests** added — one per high-severity finding (timeout-without-runtime, end()-surfaces-error, field-bypass-CSP) + the nested-for proof.
- **284 handler tests + 15 streaming tests pass** (was 283 + 13 before this round; added 1 handler test for the field-bypass + 2 streaming tests for the timeout + end() bugs).
- `cargo xtask check` ✅ green (1344s).

Force-pushed as commit `bdd5fee7` (amends the prior commit 3 in-place; commits 1 and 2 unchanged).

The two re-audit findings I noted as low-impact but worth fixing (Bug-8 / Bug-9 in the bench) confirm that even after an external review, fresh adversarial eyes find more. That's a good signal — please do another pass if you have time. The streaming and handler hot paths are the right places to keep poking at.

mohamedmansour:
Fixed in `6e3af609` (force-pushed).

## Root cause

Pre-existing race in the test, **widened** by the streaming SSR pipeline this PR introduces — not fixed-by-PR but exposed-by-PR.

The test sequence:

```ts
await page.locator('mp-category-nav').getByRole('link', { name: 'Shirts' }).first().click();
await expect(page).toHaveURL('/search/shirts');                                      // ← passes
await page.locator('mp-filter-list').getByRole('link', { name: 'Price: ...' }).click(); // ← race
```

`toHaveURL` only verifies the URL bar, which the client router updates on receipt of the partial-nav response. The filter-list components re-render asynchronously from the same response, so there's a window where:

* URL = `/search/shirts` ✓
* `mp-filter-list` DOM still has `href="/search/stickers?sort=..."` (stale)

A click in that window hits the stale link and lands on `/search/stickers?sort=price-desc`, exactly what CI saw.

Local 10×10× repro on my machine: passes 10/10. Origin/main also passes 10/10 with the same test code. The race window is below the schedulable threshold on a fast Mac M-series; on the GitHub Actions Linux runner with playwright workers contending for CPU, it lands inside the click.

The streaming SSR work in this PR doesn't directly affect the partial-nav code path (partials still go through `partial_response` → `render_partial`, returning JSON in one shot — not streaming). But the **initial page load** is now streamed, so on a constrained runner the framework's hydration / state-init can interleave differently with subsequent partial-nav DOM patches and widen the same pre-existing race.

## Fix

Made the test deterministic by waiting for the filter-list href to actually update to the new category before clicking:

```ts
await expect(page).toHaveURL('/search/shirts');
// Count-based wait (not visibility) — mp-filter-list emits both
// desktop and mobile-only variants, only one is `display`-ed in
// each project, but both share the updated href once the DOM
// patch lands.
await expect(
  page.locator('mp-filter-list a[href*="/search/shirts?sort=price-desc"]'),
).not.toHaveCount(0);
await page.locator('mp-filter-list').getByRole('link', { name: 'Price: ...' }).click();
```

Verified locally with the rebuilt branch binary: **10/10 passes**.

The remaining 9 e2e failures are the pre-existing macOS↔Linux screenshot baseline mismatches (4 commerce + 5 contact-book `toHaveScreenshot` tests) — same set noted on origin/main, unaffected by this PR.

akroshg (approved):

PR #295, opened 2026-05-16
chore: correct rust integration docs to match published crate

janechu:
## What

Two small fixes to the public `rust` integration:

1. **`crates/webui/src/lib.rs`** - re-export `RenderOptions` and `Result as HandlerResult` from `webui_handler` through the `webui` facade crate.
2. **`docs/guide/integrations/rust.md`** - use the published crate name (`microsoft-webui`) in the install block, fix trait-impl return types, and note the rename.

`Result` is renamed to `HandlerResult` on re-export so it does not collide with the existing `std::result::Result<T, E>` usage inside `crates/webui/src/lib.rs` (`build` and `build_to_disk`).

## Why

I'm integrating WebUI into an embedded host by following `docs/guide/integrations/rust.md`. Two things were blocking:

- `webui = "*"` in Cargo.toml resolves to an unrelated crate on crates.io. The Microsoft crate is published as `microsoft-webui`; the `[lib] name = "webui"` setting keeps `use webui::...` working unchanged, but the docs don't mention this. This costs a first-time adopter the same hour every time.
- The trait impls in all three framework examples reference `webui_handler::Result<()>` (the `webui_handler` crate is not the dependency the page tells you to add) and `webui::RenderOptions` (not currently re-exported). Copying the example into a fresh project does not compile.

The scope is intentionally narrow: just unblock the first-render-from-a-non-WebUI-host story that the page already promises.

## How I tested

```
cargo check -p microsoft-webui
```

(passes after the re-export)

Happy to also copy one of the framework examples into a scratch crate that depends only on `microsoft-webui` to confirm the rendered text compiles end-to-end - can add that as a doctest if you'd prefer.

## Notes

- Branch is `users/janechu/fix-rust-integration-docs` because that's the convention my host repo uses for agent-driven branches. Happy to rename to `janechu/fix-rust-integration-docs` (matching the local convention in `.github/skills/pr/SKILL.md`) if you prefer.
- Draft PR. There are 4 follow-up PRs queued behind this one addressing related friction in the Node integration docs, handler thread-safety docs, the `@microsoft/webui/platform.js` subpath export, and a fragment-rendering recipe page - opened as siblings rather than stacked.

### DIFF
--- a/crates/webui/src/lib.rs
+++ b/crates/webui/src/lib.rs
@@ -36,7 +36,8 @@ pub use webui_handler::route_handler::{
     ProtocolIndex,
 };
 pub use webui_handler::route_matcher::CompiledRouteCache;
-pub use webui_handler::{plugin::HandlerPlugin, HandlerError, ResponseWriter, WebUIHandler};
+pub use webui_handler::Result as HandlerResult;
+pub use webui_handler::{plugin::HandlerPlugin, HandlerError, RenderOptions, ResponseWriter, WebUIHandler};
 pub use webui_parser::CssStrategy;
 pub use webui_parser::DomStrategy;
 pub use webui_parser::Plugin;
--- a/crates/webui/src/lib.rs
+++ b/crates/webui/src/lib.rs
@@ -37,7 +37,9 @@ pub use webui_handler::route_handler::{
 };
 pub use webui_handler::route_matcher::CompiledRouteCache;
 pub use webui_handler::Result as HandlerResult;
-pub use webui_handler::{plugin::HandlerPlugin, HandlerError, RenderOptions, ResponseWriter, WebUIHandler};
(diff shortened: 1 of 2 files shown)


mohamedmansour (approved):

mohamedmansour (changes requested):
Please fix lint

janechu:
Thanks for the review! Fixed in e475181a — `cargo fmt --all --check` was failing because the `pub use webui_handler::{...}` re-export line exceeded rustfmt's line length once `RenderOptions` was added. Reformatted it across three lines per rustfmt. Reverified locally: `cargo fmt --all --check`, `cargo clippy --workspace -- -D warnings`, and `cargo xtask license-headers` all pass.

mohamedmansour (approved):

PR #302, opened 2026-05-16
docs: add incremental-adoption recipe page

janechu:
## Why

WebUI's existing integration recipes ([Rust](https://github.com/microsoft/webui/blob/main/docs/guide/integrations/rust.md), [Node](https://github.com/microsoft/webui/blob/main/docs/guide/integrations/node.md), [FAST 3.x](https://github.com/microsoft/webui/tree/main/examples/app/todo-fast)) describe the **destination** of an adoption: `webui build` producing `protocol.bin`, the `fast-v3` plugin emitting Declarative Shadow DOM with hydration markers, and the client using `declarativeTemplate() + observerMap() + enableHydration()` to bind to pre-rendered content.

That is rarely the **starting point** for an existing host. A team that wants to introduce a single FAST 3.x custom element into an established SSR stream typically needs to land code before the build pipeline, handler wiring, and routing are in place, so each piece can be reviewed and deployed independently.

There is currently no page that walks an adopter through a phased, reversible path. They have to infer it from the full-page recipes plus the Embedded fragments recipe (#299), the plugin docs, and the fast-element documentation, and figure out where the natural rollback boundaries are.

## What

Adds `docs/guide/integrations/incremental-adoption.md`, a focused recipe page organised as three PR-sized steps:

- **Phase 0** - ship the element with a `data-state` attribute bootstrap. No `webui build`, no `protocol.bin`, no handler. The element decodes its initial state in `connectedCallback` and renders via an imperative `html` template + `@customElement`. Goal: get the FAST class loaded behind a kill switch while SSR is wired up separately.
- **Phase 1** - add SSR with the `fast-v3` plugin. Introduces a thin wrapper template that names the element as a fragment entry (per the [embedded fragments recipe](#299)), wires `webui build --plugin fast-v3`, and switches the client to `declarativeTemplate()` + `enableHydration()`. The Phase 0 fallback stays in the file gated by `shadowRoot === null` so the kill switch can roll back instantly.
- **Phase 2** - remove the Phase 0 fallback. Drop the imperative template, the `@customElement` decorator, and the `data-state` decoding path. The `<f-template>` block becomes the single source of truth.

The page also calls out two valid earlier exits (Phase 0 as end state for JS-heavy SPAs that never need first-paint content; Phase 1 as end state when legacy fallback must be retained indefinitely) and links to the Embedded fragments recipe for the routing and module-CSS-dedup edge cases this recipe intentionally does not cover.

Adds the page to `docs/.webui-press/config.json` and `docs/guide/integrations/index.md` so it appears in the sidebar and the integrations landing page.

## Why not a separate `recipes/` directory

I considered placing this under a new `docs/guide/recipes/` section, but the [embedded fragments recipe](#299) sets the precedent that recipe-style pages live under `integrations/`. Keeping that pattern means the sidebar reads as a single ordered list rather than fragmenting recipe content across two top-level sections.

## Validation

- Plain Markdown; no code changes.
- No emdashes, endashes, or unescaped `{{` mustaches outside fenced code blocks.
- All `{{citation.x}}` references appear inside fenced code blocks where VitePress will not interpret them.
- Cross-references to other docs pages use relative links matching the conventions in `rust.md` and `fragments.md`.

## Related

- #295 (re-export `RenderOptions` from facade) - the Phase 1 Rust snippet imports `RenderOptions` from `webui`, which currently requires #295.
- #299 (embedded fragments recipe) - Phase 1 builds on the wrapper-template pattern.
- #300 (re-export hydration plugins from facade) - the Phase 1 Rust snippet imports `FastV3HydrationPlugin` from `webui`, which currently requires #300.
- #301 (warn that hydration plugins are required) - Phase 1 commits to the hydration plugin; this PR's prose assumes that warning is visible on the Plugins page.

If all of those merge before this one, the Phase 1 snippets compile as-shown against `webui` from the facade alone. If any do not, an adopter can substitute `webui_handler::plugin::fast_v3::FastV3HydrationPlugin` etc.; the recipe text still reads correctly.

### DIFF
--- a/docs/.webui-press/config.json
+++ b/docs/.webui-press/config.json
@@ -154,6 +154,10 @@
           {
             "text": "WebAssembly",
             "link": "/guide/integrations/wasm"
+          },
+          {
+            "text": "Incremental adoption",
+            "link": "/guide/integrations/incremental-adoption"
           }
         ]
       },
--- a/docs/guide/integrations/incremental-adoption.md
+++ b/docs/guide/integrations/incremental-adoption.md
@@ -1,0 +1,223 @@
+# Incremental Adoption
+
+The WebUI integration recipes assume the SSR pipeline is fully wired: `webui build` produces a `protocol.bin`, a handler plugin renders Declarative Shadow DOM with hydration markers, and the client uses `declarativeTemplate()` plus `enableHydration()` to bind to the pre-rendered content. That is the destination. It is rarely the starting point for an existing host.
+
+This page covers the typical path *to* that destination: how to ship a FAST 3.x custom element behind a kill switch in three steps, where each step is a small, reversible PR that can be deployed independently. Hosts that already adopt the [Embedded fragment](./fragments) recipe in one shot can skip directly to Phase 1; hosts that need to land code before the build pipeline is in place start at Phase 0.
+
+The running example is a `<my-citation-group>` element that renders a list of citation chips inside an existing app's response stream. Swap the tag name for whatever you are introducing.
+
+## Phase 0 - Ship the element with a data-attribute bootstrap
+
+Goal: get the FAST class loaded, registered, and rendering on the client without depending on `webui build`, `protocol.bin`, or the handler pipeline at all. This phase exists so that the element can be merged, code-reviewed, and exercised behind a feature flag while you separately wire up the SSR side.
+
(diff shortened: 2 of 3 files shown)


mohamedmansour:
I don't belive we should add this doc at all, every team has their own frameworks their own clicent side and we are overly complicating our documentation which will become stale as we continue

PR #317, opened 2026-05-29
Fix event handler args in repeat scopes

jibin7jose:


jibin7jose:
@microsoft-github-policy-service agree

mohamedmansour (approved):
We usually do not accept PRs right now since we are still in development so I will do some follow up fixes to this, thank you!

PR #330, opened 2026-06-08
fix: preserve raw style text in compiled templates

### LINKED ISSUE #319: HTML-like tag inside CSS comment corrupts client-side template `h` field (`--css style`)
## Summary

When a component's CSS file contains an HTML-like tag inside a comment
(e.g. `/* The <my-component> element ... */`), `webui build --css style`
emits a malformed `h` field for that template in
`window.__webui.templates`. The `</style>` closing tag ends up at the very
end of the string and the HTML body is trapped *inside* the `<style>`
element.

SSR (declarative shadow DOM via `<template shadowrootmode="open">`) is
unaffected and renders correctly. But `document.createElement('my-component')`
on the client produces an empty component, because the cached template HTML
is effectively just one big `<style>` whose text content is all the real
markup.

## Environment

- `@microsoft/webui` 0.0.14
- Command: `npx @microsoft/webui serve ./templates --css style --plugin webui`
- Reproduced with `curl` against the dev server (no browser needed).

## Repro

`templates/index.html`:
```html
<body>
  <h2>SSR rendered (works):</h2>
  <my-component></my-component>

  <h2>Client-side created (broken):</h2>
  <div id="dynamic"></div>

  <script>
  window.addEventListener('webui:hydration-complete', () => {
    const el = document.createElement('my-component');
    document.getElementById('dynamic').appendChild(el);
  });
  </script>
</body>
```

`templates/my-component/my-component.html`:
```html
<div class="container">
  <span>Hello from my-component</span>
</div>
```

`templates/my-component/my-component.css`:
```css
:host { display: block; }

/* The <my-component> element renders items from the data array. */
.container { padding: 16px; background: #e0e0e0; }
```

Run:
```bash
npx @microsoft/webui serve ./templates --css style --plugin webui
curl -s http://127.0.0.1:3000/
```

## Actual output (relevant snippet)

```html
<my-component><template shadowrootmode="open"><style>:host { display: block; }

/* The <my-component> element renders items from the data array. */
.container { padding: 16px; background: #e0e0e0; }</style><div class="container"><span>Hello from my-component</span></div></template></my-component>
```

That part (SSR) is fine. But the inline template registry is broken:

```js
w['my-component'] = {
  h: "<style>:host { display: block; }\n\n/* The <my-component> element renders items from the data array. */\n.container { padding: 16px; background: #e0e0e0; }</my-component><div class=\"container\">\n  <span>Hello from my-component</span>\n</div></style>",
  sd: 1
};
```

Notice:
1. The `</style>` closing tag is at the **very end** of the string, after the HTML.
2. A spurious `</my-component>` appears right after the CSS, suggesting the
   tooling fed the assembled `<style>…</style><html>` into an HTML-aware
   parser/serializer that treated `<my-component>` inside the CSS *comment*
   as a real start tag, implicitly closed `<style>`, then re-parented the
   following nodes.

## Expected

The `h` field should be equivalent to:
```
<style>…css…</style><div class="container">…</div>
```
i.e. `</style>` immediately after the CSS body, with no implicit tag
fixup performed on the CSS text. CSS inside `<style>` is raw text and
must not be HTML-parsed.

## Workaround

Remove the `<my-component>` token from the CSS comment (e.g. rephrase as
`/* The my-component element ... */`).

mohamedmansour:
## Summary

Legal CSS comments can contain HTML-like text, but the WebUI compiled-template normalizer was reparsing `<style>` contents as normal HTML. That allowed text like `<my-component>` inside preserved CSS comments to be interpreted as real markup, corrupting the client template registry `h` field by moving `</style>` after the component HTML.

This fixes the core issue by treating `<style>` contents as raw text during compiled-template normalization. Compiler-generated text-binding markers are still supported, but they are now accepted only at byte offsets recorded when the compiler emits them, so authored CSS cannot spoof `<!--t:n-->` metadata.

## Changes
- Preserve raw `<style>` text in WebUI compiled templates.
- Track generated text marker offsets to distinguish real metadata from literal CSS text.
- Add regressions for legal CSS comments with HTML-like tags, marker-like CSS text, and UTF-8/multiple signal-marker cases.

Closes #319

## Validation
- `cargo test -p microsoft-webui-parser --quiet`
- `cargo xtask check`

### DIFF
--- a/crates/webui-parser/src/plugin/webui.rs
+++ b/crates/webui-parser/src/plugin/webui.rs
@@ -219,6 +219,13 @@ impl ParserPlugin for WebUIParserPlugin {
 
 // ── Compiled template generation ───────────────────────────────────
 
+const TEXT_MARKER_PREFIX: &str = "<!--t:";
+const TEXT_MARKER_SUFFIX: &str = "-->";
+const MIN_TEXT_MARKER_DIGITS: usize = 1;
+const MIN_TEXT_MARKER_LEN: usize =
+    TEXT_MARKER_PREFIX.len() + MIN_TEXT_MARKER_DIGITS + TEXT_MARKER_SUFFIX.len();
+const TEXT_MARKER_INDEX_RADIX: usize = 10;
+
 /// A compiled template section.
 ///
 /// Used for the root template and for nested block-table entries.
@@ -1012,9 +1019,9 @@ fn compile_text_binding_at(
 
 fn emit_text_marker(meta: &mut TemplateSectionMeta, index: usize) {
     meta.text_marker_offsets.push(meta.html.len());
-    meta.html.push_str("<!--t:");
+    meta.html.push_str(TEXT_MARKER_PREFIX);
     let _ = write!(meta.html, "{index}");
-    meta.html.push_str("-->");
+    meta.html.push_str(TEXT_MARKER_SUFFIX);
 }
 
 fn compile_style_content(input: &str, meta: &mut TemplateSectionMeta) {
@@ -1560,7 +1567,7 @@ fn push_style_raw_text_nodes(
     let mut index = 0usize;
     let mut text_start = 0usize;
 
-    while index + 10 <= len {
+    while index + MIN_TEXT_MARKER_LEN <= len {
         if let Some((marker_index, marker_end)) =
             find_text_marker_comment(input, index, base_offset, text_marker_offsets)
         {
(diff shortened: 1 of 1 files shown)


crates/webui-parser/src/plugin/webui.rs
```diff
@@ -1500,6 +1549,70 @@ fn parse_fragment_nodes(input: &str) -> Vec<FragmentNode> {
...
+    input: &str,
+    base_offset: usize,
+    text_marker_offsets: &[usize],
+) {
+    let bytes = input.as_bytes();
+    let len = bytes.len();
+    let mut index = 0usize;
+    let mut text_start = 0usize;
+
+    while index + 10 <= len {
```
akroshg: what is +10 here?

akroshg (dismissed):

crates/webui-parser/src/plugin/webui.rs
mohamedmansour: changed all the marker lengths to const, thanks @akroshg please review again!

akroshg (approved):

PR #332, opened 2026-06-08
chore: add bump-version release skill

mohamedmansour:
Release bumps need a repeatable workflow that versions the workspace, validates the repo, and summarizes changes since the previous v-prefixed tag.

This Adds a local `bump-version` skill that takes a semver input, runs `cargo xtask version <version>` and `cargo xtask check`, finds the prior `v*` release tag, and formats bucketed PR notes as bulleted lists without commit SHAs.

### DIFF
--- a/.github/skills/bump-version/SKILL.md
+++ b/.github/skills/bump-version/SKILL.md
@@ -1,0 +1,142 @@
+---
+name: bump-version
+description: Bump WebUI to an input release version, run the full gate, and prepare a release PR with bucketed changes since the previous v-prefixed tag.
+---
+
+# Bump Version Release PR
+
+Use this skill when preparing a WebUI release version bump PR.
+
+## Inputs
+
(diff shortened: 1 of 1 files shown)


.github/skills/bump-version/SKILL.md
```diff
@@ -0,0 +1,142 @@
...
+description: Bump WebUI to an input release version, run the full gate, and prepare a release PR with bucketed changes since the previous v-prefixed tag.
+---
+
+# Bump Version Release PR
+
+Use this skill when preparing a WebUI release version bump PR.
+
+## Inputs
+
+Ask for the release version if the user did not provide one. The input must be the bare semver version, without a leading `v`:
```
akroshg, line 12: Does it propose a version when none provided? say increment by one?
Is there a way to have mechanism where the skill can determine what would be next version based on what we have so far in payaload from the last release? I think AI some time goes off the road, if we allow it choose the next one.

akroshg (approved):

PR #335, opened 2026-06-09
fix: scope FAST hydration markers to custom elements

janechu:
## Summary
- Scope FAST v2 and FAST v3 hydration markers to custom element component scopes.
- Keep entry page light DOM marker-free, including root signals, `for` loops, `if` blocks, and plugin element data.
- Update DESIGN.md, README.md, and plugin docs for the handler scope API and marker behavior.

## Validation
- `cargo test -p microsoft-webui-handler fast_v2 -- --nocapture`
- `cargo test -p microsoft-webui-handler fast_v3 -- --nocapture`
- `git diff --check`
- `cargo xtask check` partially completed on Windows: license headers, fmt, clippy, proto drift check, and deny passed; test phase failed on an unrelated existing Windows compile issue in `crates\webui\examples\streaming_resource_bench.rs` where `libc::rusage`, `libc::getrusage`, and `libc::RUSAGE_SELF` are unavailable.

### DIFF
--- a/crates/webui-handler/src/lib.rs
+++ b/crates/webui-handler/src/lib.rs
@@ -514,7 +514,7 @@ impl WebUIHandler {
         };
 
         if let Some(p) = &mut context.plugin {
-            p.push_scope();
+            p.push_component_scope();
         }
 
         self.process_fragment_id(entry_id, &mut context)?;
@@ -877,7 +877,7 @@ impl WebUIHandler {
         context.local_vars = saved_component_attrs;
 
         if let Some(p) = &mut context.plugin {
-            p.push_scope();
+            p.push_component_scope();
         }
 
         self.process_fragment_id(&component.fragment_id, context)?;
--- a/docs/guide/concepts/plugins/index.md
+++ b/docs/guide/concepts/plugins/index.md
@@ -121,8 +121,15 @@ pub trait ParserPlugin {
 
 ```rust
 pub trait HandlerPlugin {
-    /// Enter a new scope (component or loop item).
+    /// Enter a structural scope, such as a loop item or active if body.
     fn push_scope(&mut self);
+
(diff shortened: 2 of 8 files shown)


docs/guide/concepts/plugins/index.md
```diff
@@ -152,6 +159,14 @@ pub trait HandlerPlugin {
 }
 ```
 
+`push_component_scope` is used for component bodies and `handle_as_component`;
+`push_scope` is used for structural scopes such as loop items and active
+if-condition bodies. Plugins that need island-only markers can activate only
+inside component scopes. FAST v2/v3 use this behavior to keep normal entry page
+light DOM marker-free, including root signals, for-loops, if-conditions, and
+element metadata, while still emitting markers and data attributes inside custom
+element components and their nested structural scopes.
```
mohamedmansour, line 168: I don't believe we need this comment because  it is already implied in the method comments.

mohamedmansour (commented):
Review comments submitted.

crates/webui-handler/src/lib.rs
```diff
@@ -514,7 +514,7 @@ impl WebUIHandler {
         };
 
         if let Some(p) = &mut context.plugin {
-            p.push_scope();
+            p.push_component_scope();
```
mohamedmansour, line 517: I do not see a direct handle_as_component regression test in this PR. The current coverage checks normal full-page component rendering, but this line changes the public component-only render path as well.

Please add explicit FAST v2 and FAST v3 tests that call WebUIHandler::handle_as_component and verify hydration markers still emit inside the rendered component, including a nested structural scope such as a for-loop or if-condition. Without that, a future regression could break component-only rerender hydration while the existing full-page tests still pass.

PR #346, opened 2026-06-16
fix: move SSR state to data block

akroshg:
## Summary
- Move SSR state to the `__webui_state` JSON data block.
- Keep WebUI metadata using a lazy `window.__webui.state` getter.
- Keep FAST output data-block-only.
- Omit render-only design tokens from client bootstrap state after SSR resolves them.
- Fix the broken todo-fast app by moving delegated events into the rendered DOM and restoring the app script behavior that was tested locally.

## Benchmark
8-iteration browser benchmark, 100KB state, 4x CPU:

| Scenario | Path | Bootstrap | DCL | Load | Bytes |
|---|---|---:|---:|---:|---:|
| Inline script | buffered | 2.3 ms | 83.3 ms | 83.5 ms | 125306 |
| Data block | buffered | 1.6 ms | 80.2 ms | 80.5 ms | 125312 |
| Inline script | streaming | 2.5 ms | 80.9 ms | 81.1 ms | 125306 |
| Data block | streaming | 1.5 ms | 34.8 ms | 40.1 ms | 125312 |

## Validation
- `cargo test -p microsoft-webui-handler client_state_omit_keys_preserves_ssr_resolution`
- `cargo test -p microsoft-webui-cli test_build_and_render_omits_injected_tokens_from_client_state`
- `cargo test -p microsoft-webui-ffi`
- `pnpm --dir docs build`
- `pnpm -C examples/app/todo-fast build`
- Example app build/smoke checks
- User-tested todo-fast locally after restoring `todo-app.ts`
- `cargo xtask check` reaches existing Windows `libc::rusage/getrusage` failure in `crates/webui/examples/streaming_resource_bench.rs`

### DIFF
--- a/crates/webui-handler/src/lib.rs
+++ b/crates/webui-handler/src/lib.rs
@@ -7105,17 +7105,18 @@ mod tests {
     fn fast_v2_data_block_carries_nonce_when_csp_nonce_set() -> Result<()> {
         let protocol = fixture_route_with_body_end();
         let state = test_json!({"items": [{"title": "A"}]});
+        let nonce = std::process::id().to_string();
         let output = render_bootstrap_with_plugin(
             &protocol,
             &state,
             || Box::new(crate::plugin::fast_v2::FastV2HydrationPlugin::new()),
-            &RenderOptions::new("index.html", "/").with_nonce("test-nonce"),
+            &RenderOptions::new("index.html", "/").with_nonce(&nonce),
         )?;
 
-        assert!(output.contains(
-            "<script type=\"application/json\" id=\"__webui_state\" nonce=\"test-nonce\">"
-        ));
-        assert!(!output.contains("<script nonce=\"test-nonce\">window.__webui"));
+        assert!(output.contains(&format!(
+            "<script type=\"application/json\" id=\"__webui_state\" nonce=\"{nonce}\">"
+        )));
+        assert!(!output.contains(&format!("<script nonce=\"{nonce}\">window.__webui")));
         Ok(())
     }
 
--- a/crates/webui-handler/src/lib.rs
+++ b/crates/webui-handler/src/lib.rs
@@ -259,13 +259,14 @@ struct WebUIProcessContext<'a> {
 }
 
 struct WebUiBootstrap<'a> {
-    state: &'a Value,
     chain: &'a [Value],
     inventory: &'a str,
     nonce: Option<&'a str>,
     css_hrefs: &'a [&'a str],
     style_specs: &'a [&'a str],
 }
+
+const STATE_DATA_BLOCK_ID: &str = "__webui_state";
 
 /// Get the component attribute name, stripping `:` prefix and converting to camelCase.
 ///
@@ -357,7 +358,7 @@ where
     write_script_safe_json(writer, value)
 }
 
-fn write_webui_bootstrap(
+fn write_webui_metadata_init(
     writer: &mut dyn ResponseWriter,
     bootstrap: WebUiBootstrap<'_>,
 ) -> Result<()> {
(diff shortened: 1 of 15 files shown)


crates/webui-handler/src/plugin/fast_v2.rs
```diff
@@ -197,6 +197,10 @@ impl HandlerPlugin for FastV2HydrationPlugin {
     ) -> Result<()> {
         write_fast_route_component_state(state, writer)
     }
+
+    fn needs_webui_bootstrap(&self) -> bool {
```
mohamedmansour, line 201: What is this about? Why do we need a plugin for just this? "webui" is a plugin why does plugin metadata need to know about a specific plugin

crates/webui-handler/src/plugin/mod.rs
```diff
@@ -114,6 +114,14 @@ pub trait HandlerPlugin {
     ) -> Option<Vec<String>> {
         None
     }
+
+    /// Whether this plugin needs the `window.__webui` metadata bootstrap
+    /// (`chain`, `inventory`, `nonce`, `css`, `styles`, `templates`).
```
mohamedmansour, line 119: chain/inventory are all part of the webui-router package part of the framework

crates/webui-handler/src/lib.rs
```diff
@@ -383,6 +383,35 @@ fn write_webui_bootstrap(
...
+        writer.write(" nonce=\"")?;
+        writer.write(nonce)?;
+        writer.write("\"")?;
+    }
+    writer.write(">")?;
+    write_script_safe_json(writer, state)?;
+    writer.write("</script>")
+}
+
+fn write_lazy_state_getter(writer: &mut dyn ResponseWriter) -> Result<()> {
```
mohamedmansour, line 457: Did the "webui" plugin regress or improve with this change?

examples/app/todo-fast/src/todo-item/todo-item.html
```diff
@@ -1,13 +1,11 @@
...
-      <if condition="state == 'done'">
-        <span class="check">&#10003;</span>
-      </if>
-    </button>
-    <span class="title">{{title}}</span>
-    <button class="delete" data-action="delete" title="Delete">&times;</button>
-  </div>
-</template>
+<template shadowrootmode="open">
+  <div class="item" @click="{onClick($e)}">
```
mohamedmansour, line 2: Why did this move to a div tag? The whole point is the shadowroot @click events wont work (since FAST does this, so we support it)

examples/app/todo-fast/src/todo-app/todo-app.ts
```diff
@@ -34,9 +36,14 @@ export class TodoApp extends FASTElement {
     console.log('TodoApp disconnected');
   }
 
-  private prepareFromDom(): void {
+  private prepareOnce(): void {
```
mohamedmansour: Why would this be called more than once?

crates/webui-handler/src/lib.rs
```diff
@@ -7029,6 +7039,128 @@
...
+
+    #[test]
+    fn fast_v2_data_block_carries_nonce_when_csp_nonce_set() -> Result<()> {
+        let protocol = fixture_route_with_body_end();
+        let state = test_json!({"items": [{"title": "A"}]});
+        let output = render_bootstrap_with_plugin(
+            &protocol,
+            &state,
+            || Box::new(crate::plugin::fast_v2::FastV2HydrationPlugin::new()),
+            &RenderOptions::new("index.html", "/").with_nonce("test-nonce"),
```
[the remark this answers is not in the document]
akroshg: **AI Response: Fixed**

Removed the hard-coded nonce fixture from the new test. The test now uses a dynamic nonce value and still asserts that the data-block script receives the nonce while FAST does not emit a WebUI bootstrap script.

examples/app/todo-fast/src/todo-app/todo-app.ts
akroshg: Not needed - reverted.

examples/app/todo-fast/src/todo-item/todo-item.html
akroshg: This is to fix the todo-fast. I'll work with Jane on this.

crates/webui-handler/src/lib.rs
akroshg: Good point - let's see if I can find or create some benchmark for this.

crates/webui-handler/src/plugin/fast_v2.rs
akroshg: This for adding anonymous function to parse the state. In the fast side we will parse the state at InitialState.js. but that something will not work for `webui` plugin. Let's see if this is a good approach.

mohamedmansour:
Closing since #350

PR #356, opened 2026-06-18
fix: resolve npm CLI platform binary

janechu:
## Summary
- Replace the root @microsoft/webui native bin with a JavaScript launcher that resolves the current platform package at runtime.
- Add platform-suffixed bin entries so npm publishes platform package binaries with executable permissions.
- Update publish staging, docs, and tests to cover launcher dispatch and executable binary staging.

## Validation
- pnpm --filter @microsoft/webui test
- cargo test -p xtask stage_file_sets_executable_permissions_when_requested
- npm tarball checks confirmed root bin packs the JS launcher and macOS platform bin packs with executable mode.
- cargo xtask check

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,4 @@ packages/*/webui.node
 publish/
 *.nupkg
 .last
-packages/webui/bin/webui
 packages/webui/bin/webui.exe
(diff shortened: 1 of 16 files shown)


mohamedmansour (changes requested):
We don't want to have a JS middleware to launch the rust cli, that is against the point, we basically copy the native library to the bin so that it directly goes to native instead of proxying the hop through JS

PR #358, opened 2026-06-19
fix: preserve root runtime attrs for plugin templates

akroshg:
## Summary
- Fix root `<template>` runtime attrs disappearing from plugin client templates.
- Keep SSR output clean by stripping runtime attrs there.
- Preserve `@`, `:`, `?`, and FAST `f-*` attrs only for plugin-facing templates.

## Why
PR #307 made the parser preserve authored `<template>` wrappers. That is correct for SSR, but runtime attrs on the wrapper were stripped before FAST received the client template artifact. Apps with root event handlers, like todo-fast and Edge Downloads BTR, rendered correctly but did not wire those handlers.

This fixes the split at the parser boundary: SSR gets the stripped template, while plugins get the client artifact with root runtime attrs preserved. That is better than changing apps or adding FAST-specific restoration logic.

## Validation
- `cargo test -p microsoft-webui-parser plugin_artifact_template_keeps_root_runtime_attrs`
- `cargo test -p microsoft-webui-parser component_template_strips_shadowrootmode_from_f_template`
- `cargo test -p microsoft-webui-parser fast_v3`
- `cargo test -p microsoft-webui-parser fast_v2`
- `pnpm -C examples/app/todo-fast build`
- `pnpm --dir docs build`
- `cargo xtask check` reaches the existing Windows `libc::rusage/getrusage` failure in `crates/webui/examples/streaming_resource_bench.rs`

### DIFF
--- a/crates/webui-parser/src/lib.rs
+++ b/crates/webui-parser/src/lib.rs
@@ -28,6 +28,7 @@ pub use handlebars_parser::HandlebarsParser;
 
 use crate::html_parser::{self as html, Attrs, Element, Event, Walker};
 use crate::plugin::{AttributeAction, ParserPlugin, ParserPluginArtifacts};
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::ops::Range;
 use webui_protocol::{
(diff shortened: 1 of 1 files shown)


akroshg:
#359 has a better fix.

akroshg:
Closing in favor of #359.

PR #359, opened 2026-06-19
fix: preserve plugin template root attributes

mohamedmansour:
## Summary

- Preserve user-authored root `<template>` attributes for parser plugin/client component templates while continuing to strip runtime-only attributes from the SSR/internal parse view.
- Build explicit component template views in the parser: an SSR view for fragment parsing and a plugin/client artifact view for `ParserPlugin::register_component_template`.
- Append `shadowrootadoptedstylesheets="<component-name>"` centrally for CSS module strategy when a component has CSS and the authored wrapper is missing the attribute.
- Remove the obsolete `ParserError::MissingAdoptedStylesheets` contract and update `DESIGN.md`, CLI docs, plugin docs, and the AI reference.
- Add `todo-fast` Playwright coverage and wire it into `cargo xtask e2e` so FAST delegated root events are covered without requiring CSS module mode.

## Why

FAST component templates were receiving the SSR-cleaned component template string. That string had root runtime attributes such as `@toggle-item`, `@delete-item`, and `@click` stripped from user-authored `<template>` wrappers before FAST generated `<f-template>` artifacts. As a result, `examples/app/todo-fast` rendered and could handle local child events, but delegated root events from child components did not wire up, so toggling and deleting existing todos failed.

The correct contract is that authored root `<template>` attributes belong to the plugin/client template surface. SSR can still use a clean view that removes runtime-only syntax, but plugins should not have to reconstruct or special-case raw component source.

## Design

This change centralizes the behavior in `HtmlParser` rather than adding a FAST-specific workaround:

- `BuiltComponentTemplate::ssr` is used for internal fragment parsing and SSR output.
- `BuiltComponentTemplate::artifact()` is passed to parser plugins.
- The artifact view preserves user-authored root `<template>` attributes.
- The SSR view strips runtime-only root attrs (`@`, `:`, `?`) so rendered HTML remains clean.
- CSS module adopted stylesheet attrs are appended in the shared builder when missing, and authored multi-specifier values are preserved verbatim.

This keeps WebUI and FAST on the same parser/plugin contract. WebUI’s existing raw-source root-event path remains compatible, but plugin correctness no longer depends on duplicating that pattern in FAST.

## Performance notes

- No regex or recursion added.
- The second artifact template is only built when a plugin exists and authored root runtime attrs would otherwise be stripped.
- Existing quote-aware scanner helpers are reused for template tag and attribute handling.
- Buffers are pre-sized for CSS injection and adopted stylesheet insertion.

## Validation

- `cargo test -p microsoft-webui-parser --lib`
- `cargo test -p microsoft-webui-parser`
- `pnpm --dir examples/app/todo-fast run build`
- `pnpm --dir examples/app/todo-fast test`
- `pnpm --dir examples/app/todo-webui test`
- `pnpm --dir docs build`
- `cargo xtask check`


akroshg (approved):

crates/webui-parser/src/lib.rs
```diff
@@ -2038,6 +2051,7 @@ impl HtmlParser {
     const SKIPPED_ATTRIBUTES: &[&str] = &["class", "style", "role"];
     /// Skipped attribute prefixes for components.
     const SKIPPED_ATTRIBUTE_PREFIXES: &[&str] = &["data-", "aria-"];
+    const ADOPTED_STYLESHEETS_ATTR: &str = "shadowrootadoptedstylesheets";
```
janechu, line 2054: similar to `SKIPPED_ATTRIBUTE_PREFIXES` should `ADOPTED_STYLESHEETS_ATTR` target something like `SHADOWROOT_ATTRIBUTE_PREFIXES`, this way you could include `shadowrootdelegatesfocus` and future-proof against other `shadowroot` prefixed attributes being applied.

janechu (approved):

crates/webui-parser/src/lib.rs
mohamedmansour: shadowrootadoptedstylesheets  is singled out because it is the only shadowroot attr WebUI conditionally appends for  --css=module ; other  shadowroot*  attrs are user-authored and  preserved, not generated.

PR #360, opened 2026-06-19
feat: add ESM static component assets

mohamedmansour:
## Summary

Adds ESM static component assets so WebUI apps can keep the initial SSR payload small while loading deferred component trees asynchronously from a CDN or static folder without `@microsoft/webui-router` and without a template server.

`webui build --emit-component-assets <tags>` compiles requested lazy roots through synthetic non-entry fragments, keeping them out of the initial SSR entry graph while still emitting their full component dependency closure as static `.webui.js` modules.

## What changed

- Emit one browser-native ESM asset per requested root, for example `lazy-panel.webui.js`.
  - The module default-exports template metadata, CSS module import maps, dependency closure metadata, and compiled condition functions together.
  - Static assets intentionally omit inventory state because they cannot know the page's current loaded template bitset.
- Add `@microsoft/webui-framework/component-asset.js` with `defineComponentAssets()`.
  - Starts asset, component module, and optional data work together.
  - Skips importing when the root template is already registered.
  - Deduplicates in-flight imports and CSS module import maps.
  - Creates elements after asset/module work is ready and applies data later via `setState()` by default, with optional bounded data blocking.
- Move static component asset rendering into the Rust core API.
  - `BuildOptions::component_asset_roots` requests asset roots.
  - `BuildResult::component_asset_files` returns rendered `[filename, ESM content]` files.
  - `render_component_assets()` exposes the renderer for Rust callers.
  - `build_to_disk()` writes component assets alongside `protocol.bin` and CSS files.
- Expose the same build surface through Node/npm.
  - Node native builds accept `componentAssetRoots`.
  - `build()` returns flattened `componentAssetFiles` (`[filename, content, ...]`) in addition to `cssFiles`.
- Add generic emitted asset filename templating.
  - Reuses `[name]`, `[hash]`, `[ext]` for Link-mode CSS files and static component assets.
  - `[ext]` resolves to `css` for CSS and `webui.js` for component assets.
  - Protocol, CSS, and component asset filenames are validated together before writes, preventing partial output and overwrite collisions.
- Keep static `.webui.js` modules out of preload/link handling; only CSS assets use the existing CSS preload/link paths.
- Remove synthetic asset-root wrapper fragments before protocol serialization so asset-only wrappers do not bloat `protocol.bin`.
- Add `examples/app/component-assets` showing a no-router, no-template-server deferred component flow.
- Update DESIGN.md, CLI docs, Rust/Node integration docs, AI docs, interactivity docs, package READMEs, and framework docs.

## Follow-ups filed

- #364 — Support scoped static component asset emission for incremental/HMR and bundler-plugin flows.
- #365 — Deduplicate shared dependencies across static component assets.

## Validation

- `cargo test -p microsoft-webui -p microsoft-webui-cli --lib`
- `cargo test -p microsoft-webui-cli`
- `cargo test -p microsoft-webui-node --lib`
- `cd packages/webui && pnpm build && pnpm test`
- `cargo xtask fmt`
- `cargo xtask clippy`


Qusic (dismissed):

crates/webui/src/component_assets.rs
```diff
@@ -0,0 +1,515 @@
...
+/// conservative component dependency closure, template/style metadata, and any
+/// WebUI condition closures needed by those templates.
+///
+/// # Errors
+///
+/// Returns [`WebUIError`] when the root allowlist is invalid, a requested root
+/// has no compiled template metadata, asset filename generation fails, or two
+/// component assets resolve to the same filename.
+#[must_use = "component asset files must be written or otherwise consumed"]
+pub fn render_component_assets(
```
Qusic, line 43: non-blocking / extensibility: render_component_assets(protocol, …) taking a full protocol means emitting a single component asset still requires a full webui::build() (entry parse + full protocol.bin). For a dev/HMR or incremental bundler-plugin flow, you'd want to re-emit just one changed root without rebuilding the whole app.

Not a blocker. Future idea: a scoped entry that parses only the requested roots + their closure into a minimal protocol (basically the synthetic-non-entry-fragment path already in parse_component_asset_roots, minus parsing the app entry), so this can run per-component on watch.

crates/webui/src/component_assets.rs
```diff
@@ -0,0 +1,515 @@
...
+        && bytes
+            .iter()
+            .all(|b| b.is_ascii_lowercase() || b.is_ascii_digit() || *b == b'-')
+}
+
+fn has_template_payload(component: &webui_protocol::ComponentData) -> bool {
+    !component.template_json.is_empty() || !component.template.is_empty()
+}
+
+fn collect_component_asset_closure(protocol: &WebUIProtocol, root: &str) -> Vec<String> {
```
Qusic, line 339: non-blocking / future consideration: Each root inlines its full dependency closure into one self-contained module, so a component shared by multiple roots (or already in the initial SSR bundle) gets duplicated byte-for-byte across every asset that reaches it. Native ESM module-graph dedup can't help here because the modules embed the closure as a literal rather than import-ing the dependency component modules.

Not a blocker for this PR — the runtime dedups registration via getTemplate(tag) / templatesAlreadyRegistered, so correctness is fine. Flagging for later: emitting one module per component that cross-imports its dependencies (or factoring shared deps into a common chunk) would let the browser / bundler dedup the bytes too, which matters once a design-system component is shared across many lazy roots.

crates/webui/src/lib.rs
```diff
@@ -91,6 +104,7 @@ impl Default for BuildOptions {
             dom: DomStrategy::Shadow,
             plugin: None,
             components: Vec::new(),
+            component_asset_roots: Vec::new(),
```
Qusic, line 107: non-blocking / extensibility: The Rust side now exposes a clean webui::render_component_assets() + BuildResult.component_asset_files — nice. The Node binding, though, hardcodes component_asset_roots: Vec::new() and JsBuildResult only surfaces css_files, so JS tooling (a Vite/webpack plugin) can't request roots or read the emitted ESM modules in-process — it has to shell out to the CLI and glob dist.

Not needed for this PR. Flagging for when we want a first-class bundler plugin: threading component_asset_roots through JsBuildOptions and adding component_asset_files (name + content) to JsBuildResult would let a plugin run emit in-process and feed the modules straight into the bundler graph (emitFile/resolveId), instead of shell-out + read-back.

mohamedmansour: Fixed in f6015914: Node build options now accept componentAssetRoots, native build results return flattened componentAssetFiles, and the npm package TypeScript API/docs/tests reflect the new result field.

mohamedmansour: Done

crates/webui/src/component_assets.rs
mohamedmansour: Filed #365

crates/webui/src/component_assets.rs
mohamedmansour: Filed #364

Qusic (approved):

akroshg (approved):

PR #367, opened 2026-06-24
feat: add lean Rust desktop SDK with typed IPC

mohamedmansour:
## Summary

- Add one Rust-native desktop SDK using WKWebView on macOS, WebView2 on Windows, and GTK4/WebKitGTK 6 on Linux, without Electron, a bundled browser engine, Node, or a localhost application server.
- Keep production apps lean: `microsoft-webui-desktop` has no default features. `native` enables the system webview; `source` enables development compilation; `cli` enables separate tooling. App-root packaging defaults to optimized runtime-only builds.
- Use the same `DesktopApp` builder for source and immutable bundles, preserving window configuration, shell settings, application identity, route providers, and handlers before startup rendering.
- Include standalone app scaffolding, the Contact Book Rust host, native/portable packaging, and platform-neutral window customization with explicit capability validation.

## Typed application IPC

Replace the unmerged string-method API with generated Rust and TypeScript interfaces from one proto3 contract:

| Flow | Application API |
|---|---|
| JavaScript to Rust request/reply | Generated host methods returning typed promises |
| Rust to JavaScript request/reply | Generated renderer clients returning typed futures |
| Rust to JavaScript notifications | Generated emitters and scoped JS subscriptions |
| JavaScript to Rust notifications | Generated emitters and Rust startup/session handlers |

- Add `webui desktop ipc generate`, the build-time `microsoft-webui-desktop-build` crate, and browser-only `@microsoft/webui-desktop`. Shared bindings preserve bigint/64-bit values, binary bytes, native maps, optional fields, and oneofs.
- Enforce schema identity, stable method IDs and retirement history, generated validation, and explicit host permissions. Generation checks output containment, stages changes with rollback on reported publication failures, and supports drift-only `--check`.
- Keep application payloads protobuf binary. Native messages carry bounded admission and availability metadata, without base64 payloads, per-byte object arrays, idle polling, or automatic retries.
- Run Rust application handlers off the native UI thread on lazy, bounded SDK workers. Requests support deadlines and cooperative cancellation; notification acceptance does not imply subscriber completion.
- Bound SDK-owned frames, bytes, pending calls, tasks, and callbacks. Preserve byte reservations through actual native NSData, GBytes, and IStream lifetimes; retain separate control capacity under payload saturation.
- Bind credentials, callbacks, and handles to an admitted document. Full navigation and close retire old sessions; same-document routing preserves them. BFCache restoration explicitly reconnects through fresh native admission, without reviving old handles or replaying calls.
- Serve the same embedded bootstrap/runtime in source and packaged modes. Lifecycle events and window controls remain separate from application IPC.

## Window customization and ownership

Cross-platform options cover native/custom titlebars, caption-control insets, background, effects, geometry, and native controls. Unsupported features fail before launch rather than silently becoming no-ops. Frames own lifecycle subscriptions, bounded command queues, and IPC state; weak handles cannot keep a closed frame accepting work. Wakeups are coalesced, callbacks run outside registry locks, and persisted geometry uses bounded reads and atomic replacement.

## Release resource measurements

Measured SDK source: `b72f4e91`, unchanged by the subsequent CI metadata repair. Apple M4 Pro, macOS arm64, optimized runtime-only builds, one warmup per mode and five rotated triples. This is a **minimal empty-schema harness, not the Contact Book application**. Registered-unused injects the bootstrap; admitted-idle additionally imports the real browser runtime and retains an authenticated connection without application traffic.

| Median or artifact | IPC disabled | Registered-unused | Admitted-idle |
|---|---:|---:|---:|
| Host physical footprint | 22.079 MiB | 22.032 MiB | 22.079 MiB |
| WebContent physical footprint | 14.719 MiB | 15.032 MiB | 15.813 MiB |
| Executable | 1,475,424 B | 1,475,520 B | 1,475,520 B |
| Executable + logical resources | 1,524,979 B | 1,525,075 B | 1,525,075 B |

Admission versus registered-unused adds a paired median **0.781 MiB** of WebContent footprint (range +0.641 to +0.984 MiB). The paired host delta, +0.062 MiB, is within observed variability; GPU variation prevents a strong total-memory conclusion. Both executable variants have 49 normal-edge dependency entries excluding the harness, without compiler/parser, Tokio, Rayon, prost-build or prost-reflect. Disabled builds still include shared SDK IPC code, so the 96-byte executable difference is not the total size of the IPC implementation.

Hidden process-launch-to-connection-ready p50/p95 was **254.34/358.45 ms**. Admission after DOM-ready was **2.95/52.52 ms**, retaining both slower samples. These warm-cache, hidden-window observations are not foreground startup or first-paint measurements, and no startup improvement is claimed. Product byte totals exclude observer/container metadata and signing; resource observations include constant measurement scaffolding. All 72 owned host/helper PIDs exited.

Earlier, before typed IPC, SDK consolidation reduced the measured macOS Contact Book executable from 2,494,672 to 1,314,368 bytes and its logical app bundle from 2,746,433 to 1,566,129 bytes. Those remain historical consolidation measurements, not the current Contact Book artifact size.

## Verification and remaining platform acceptance

- `cargo xtask check` passes with all new files tracked, including formatting, linting, dependency audit, workspace tests, generated-artifact checks, native/source/CLI coverage, builds, WASM, examples, benchmarks, and docs. Strict desktop native library/test Clippy also passes.
- Portable IPC, browser/runtime, and generator regressions pass. Independent reviews verified fixes for ordering, control-credit starvation, wire validation, artifact publication, native completion wakeups, negotiated Windows limits, and native/browser buffer ownership.
- Six actual WKWebView executions passed across three predetermined source/package pairs. Each verified the four flows, payloads through 256 KiB, cancellation/recovery, same-document routing, explicit close, and two BFCache restorations with old callbacks/handles remaining inactive. No hidden errors, retries, or forced cleanup were accepted.
- macOS navigation testing reproduced and fixed browser-owned outbound GET cancellation before page retirement. Bodyless reads remain explicitly abortable by the SDK; payload POSTs do not opt into the browser's 64 KiB keepalive quota, and genuine transport failures remain errors. Independent delayed-response and partial-stream probes verify replacement isolation and credit release.
- The first new CI run stopped during npm supply-chain verification, before native testing. Eight machine-local mirror URLs were removed through lockfile regeneration with **all versions and integrity pins unchanged**; strict verification remains enabled. The existing generated-file license exemption now explicitly includes the two protoc-generated browser codecs, with authored runtime/test checks retained.
- Windows production cross-compilation/Clippy and Linux binding checks pass locally. CI runs the actual source/package fixture on all three operating systems and retains diagnostic artifacts. **Live Windows and Linux native acceptance is pending the repaired CI run; local cross-checking is not runtime proof.**

The previous macOS Contact Book UI/screenshot verification remains applicable to the unchanged visual shell; the new IPC fixture is non-visual protocol verification.


crates/webui-desktop/src/event.rs
```diff
@@ -0,0 +1,403 @@
...
+            handlers.push(Arc::new(handler));
+        }
+    }
+    /// Dispatch an event and report cancellation.
+    #[must_use]
+    pub fn dispatch(&self, event: &DesktopEvent) -> EventResponse {
+        let Ok(handlers) = self.handlers.lock() else {
+            return EventResponse::Continue;
+        };
+        if handlers
```
mddinbox: `dispatch` still holds `handlers`' mutex while invoking arbitrary user callbacks. If a callback calls `on_event` or dispatches another event, it tries to acquire the same non-reentrant mutex and deadlocks the UI thread. Clone the handler `Arc`s under the lock, release it, then invoke the snapshot so callbacks can safely register or dispatch.

crates/webui-desktop/src/ipc.rs
```diff
@@ -0,0 +1,366 @@
...
+            request_id,
+            result: Some(desktop_ipc_response::Result::Error(DesktopIpcError {
+                code: code.into(),
+                message: message.into(),
+                help: help.into(),
+            })),
+        })
+    }
+
+    fn encode_response(&self, response: DesktopIpcResponse) -> Result<Vec<u8>> {
```
mddinbox: A registered handler's response is encoded here without applying `max_payload_bytes`; the limit is enforced only on the inbound frame and request payload. A small browser request can therefore produce an arbitrarily large response allocation/transport, contrary to the documented 1 MiB IPC frame cap and allowing memory/IPC exhaustion. Reject oversized handler payloads and the final encoded response frame with the existing structured `payload-too-large` error before returning it.

crates/webui-desktop/src/event.rs
mohamedmansour: Good catch - fixed. The handler list is now copy-on-write behind `Arc<[Arc<EventHandler>]>`; `dispatch` clones the `Arc` and drops the guard before invoking anything, so a callback can register or dispatch freely. Cloning is a refcount bump rather than a per-dispatch allocation, which matters since resize events fire continuously during a drag.

While in there I also stopped `dispatch` short-circuiting on the first `PreventDefault` - it used `any()`, so a cancelling handler hid the event from every handler registered after it, making delivery depend on registration order. All handlers now run and the response is still `PreventDefault` if any cancelled.

Added two regression tests. The deadlock one dispatches on a worker thread with a bounded wait so a regression fails the suite instead of hanging it forever; I verified it fails when the guard is held across callbacks.

crates/webui-desktop/src/ipc.rs
mohamedmansour: Agreed, fixed. The response is now bounded in two places: the handler's returned payload is checked before encoding, so an oversized result never gets copied into a response buffer just to be rejected, and the finished frame is checked after encoding, since protobuf framing and a long handler error message both add bytes on top of the payload. Both return the existing structured `payload-too-large` error.

One thing worth recording: writing the test surfaced that the rejection frame is itself a couple hundred bytes, so a cap configured below that floor cannot be honored. I return that frame without re-checking it rather than recursing, and documented the floor on the helper - the default is 1 MiB, so it only matters for a misconfigured limit.

Two regression tests cover it: an oversized payload, and a payload exactly at the cap where only the framing pushes it over. Both fail without the fix.

mddinbox (dismissed):
Reviewed the current head c291ddf7 and rechecked both prior findings after the author fixes. Event dispatch now releases the registry lock before callbacks, and IPC response payloads and encoded frames are bounded with structured errors. The focused desktop suite passes all 58 tests and CI is green. No remaining issues found.

mddinbox (dismissed):
Reviewed the latest head a9a607d4, including the delta after c291ddf7 and the affected desktop runtime, packaging, platform, CI, and documentation contracts. Rechecked both previously authored threads: the event registry lock/lifetime fix and IPC response/frame bounds are present, and both threads are resolved. Focused validation passed on Windows: cargo test -p microsoft-webui-desktop --no-default-features --features native (77 tests), cargo test -p microsoft-webui-desktop --no-default-features (63 tests), and cargo clippy -p microsoft-webui-desktop --no-default-features -- -D warnings. No remaining confirmed issues.

.github/workflows/pr.yml
```diff
@@ -61,6 +85,28 @@ jobs:
...
 
+      - name: Install native IPC test display
+        run: sudo apt-get install -y -qq dbus-x11 xvfb xauth
+
+      - name: Fetch native IPC fixture dependencies
+        run: cargo fetch --locked --manifest-path crates/webui-desktop/tests/fixtures/native-ipc/Cargo.toml
+
+      - name: Test native IPC source and package
+        run: |
+          GDK_BACKEND=x11 dbus-run-session -- xvfb-run -a -s "-screen 0 1280x800x24" \
```
mddinbox: The current Linux job starts dbus-run-session before xvfb-run, so the D-Bus session daemon and services it launches do not inherit the Xvfb DISPLAY. At this head the fixture logs repeated xdg-desktop-portal-gtk: cannot open display warnings and both source and packaged runs exit without a NATIVE_IPC_RESULT, making the new required CI test fail before validating IPC. Start Xvfb outside the D-Bus session (for example, xvfb-run ... dbus-run-session -- python3 ...) or explicitly export the X display into the D-Bus environment.

mddinbox (changes requested):
The current Linux job starts dbus-run-session before xvfb-run, so the D-Bus session daemon and services it launches do not inherit the Xvfb DISPLAY. At this head the fixture logs repeated xdg-desktop-portal-gtk: cannot open display warnings and both source and packaged runs exit without a NATIVE_IPC_RESULT, making the new required CI test fail before validating IPC. Start Xvfb outside the D-Bus session (for example, xvfb-run ... dbus-run-session -- python3 ...) or explicitly export the X display into the D-Bus environment.

crates/webui-desktop/src/linux/ipc_control.rs
```diff
@@ -0,0 +1,190 @@
...
+        return None;
+    }
+    value.parse().ok()
+}
+
+fn hex(value: &str) -> Option<[u8; 16]> {
+    if value.len() != 32 {
+        return None;
+    }
+    let mut result = [0; 16];
```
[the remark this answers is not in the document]
mohamedmansour: Declined as a false positive. The zeroed array is only the destination buffer for strict 32-character hex decoding, and every byte is overwritten before it can be returned; the nonce itself is generated in the renderer with `crypto.getRandomValues` and verified during admission.

.github/workflows/pr.yml
mohamedmansour: Accepted the launcher ordering change, but the next CI run exposed a separate Ubuntu 24.04 AppArmor denial of bubblewrap UID mapping. The workflow now also loads a fixture-scoped user-namespace profile for the test and removes it through an exit trap, while keeping WebKit's sandbox and the global AppArmor restriction enabled.

.github/workflows/pr.yml
```diff
@@ -61,6 +85,33 @@ jobs:
...
+
+      - name: Test native IPC source and package
+        run: |
+          # Allow only fixture executables to create WebKitGTK sandbox namespaces.
+          # Keep both AppArmor's global restriction and WebKit's sandbox enabled.
+          profile=crates/webui-desktop/tests/fixtures/native-ipc/linux-apparmor
+          sudo apparmor_parser -r "$profile"
+          trap 'sudo apparmor_parser -R "$profile"' EXIT
+          GDK_BACKEND=x11 xvfb-run -a -s "-screen 0 1280x800x24" dbus-run-session -- \
+            python3 crates/webui-desktop/tests/fixtures/native-ipc/run.py --timeout 60
```
mddinbox, line 114: This new Linux merge-gating step currently fails in both source and packaged modes with `NATIVE_IPC_DIAGNOSTIC failure code=transport`, producing no `NATIVE_IPC_RESULT` (run 35683841343). The launcher/AppArmor ordering is no longer the observed failure; the current native IPC transport still cannot complete. Please fix the Linux transport/sandbox contract and rerun this fixture before merging.

.github/workflows/pr.yml
```diff
@@ -83,6 +156,26 @@ jobs:
...
         run: cargo test -p microsoft-webui-dev-server -p microsoft-webui-cli -p microsoft-webui-press
 
+      - name: Test native desktop SDK
+        run: cargo test -p microsoft-webui-desktop --no-default-features --features native
+
+      - name: Fetch native IPC fixture dependencies
+        run: cargo fetch --locked --manifest-path crates/webui-desktop/tests/fixtures/native-ipc/Cargo.toml
+
+      - name: Test native IPC source and package
+        run: python crates/webui-desktop/tests/fixtures/native-ipc/run.py --timeout 60
```
mddinbox, line 204: This new Windows merge-gating step currently fails in both source and packaged modes during `history-forward` with `code=not-ready` (run 35683841343), so the fixture exits nonzero. That path is required to succeed; only the intentionally pending request path accepts navigation/close rejection. Please synchronize admission with the native readiness event or fix the readiness ordering rather than accepting this error.

mddinbox (changes requested):
The newly required native IPC checks are still red on this head, so this PR is not ready to merge. Linux fails in both source and packaged modes with `NATIVE_IPC_DIAGNOSTIC failure code=transport` and no `NATIVE_IPC_RESULT`; Windows fails in both modes during `history-forward` with `code=not-ready`. Please fix the platform readiness/transport issues and rerun the fixture on all three native targets rather than weakening the assertions.

.github/workflows/pr.yml
mddinbox: Verified on the current head `81240e3f`: the native IPC readiness/retirement changes cover the history-forward path, and the latest PR Checks run `35713468253` passed Build (Windows) and the native IPC fixture. The prior `not-ready` failure is resolved.

.github/workflows/pr.yml
mddinbox: Verified on the current head `81240e3f`: the Linux launcher now runs `GDK_BACKEND=x11 xvfb-run ... dbus-run-session -- python3 ...`, and the fixture-scoped AppArmor profile is loaded and removed around the run. The latest PR Checks run `35713468253` passed Build (Linux) and the native IPC fixture, so the transport failure is resolved.

mddinbox (dismissed):
Re-reviewed the current head 81240e3f after the fixes. The previously reported Linux transport and Windows readiness failures are resolved, both authored threads are resolved, and the latest PR Checks run 35713468253 passed all jobs including Build (Linux), Build (Windows), and the native IPC fixture. No remaining confirmed issues.

packages/webui-router/src/fallback-navigation.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  let url: URL;
+  let current: URL;
+  try {
+    current = new URL(currentHref);
+    url = new URL(link.resolved, currentHref);
+  } catch {
+    return null;
+  }
+  if (url.origin !== current.origin) return null;
```
mddinbox: For non-HTTP schemes, `URL.origin` serializes to the opaque value `null` for different authorities. Therefore a link from `webui://app/...` to `webui://other/...` passes this check and is intercepted as an in-app navigation, even though it is not the same custom-scheme authority. That can route the app to another registered `webui://` authority instead of leaving the navigation to the host. Compare the protocol and authority components (while preserving the normal HTTP(S) same-origin behavior), and add a regression test for `webui://app` -> `webui://other`.

mddinbox: For non-HTTP schemes, `URL.origin` serializes to the opaque value `null` for different authorities. Therefore a link from `webui://app/...` to `webui://other/...` passes this check and is intercepted as an in-app navigation, even though it is not the same custom-scheme authority. That can route the app to another registered `webui://` authority instead of leaving the navigation to the host. Compare the protocol and authority components (while preserving the normal HTTP(S) same-origin behavior), and add a regression test for `webui://app` -> `webui://other`.

mohamedmansour: Fixed in 40e62484. Custom-scheme navigation now compares protocol and authority components when URL.origin is opaque, and the regression suite covers webui://app -> webui://other so that link is left to the host.

mohamedmansour: Fixed in 40e62484. Custom-scheme navigation now compares protocol and authority components when URL.origin is opaque, and the regression suite covers webui://app -> webui://other so that link is left to the host.

pnpm-lock.yaml
```diff
@@ -1339,6 +1373,16 @@ packages:
...
     engines: {node: '>=0.6'}
 
+  ts-poet@6.12.0:
+    resolution: {integrity: sha1-8o65F3j9/44bJkYAx93dtAzTnYQ=, tarball: https://ms-feed-25.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/ts-poet/-/ts-poet-6.12.0.tgz}
+
+  ts-proto-descriptors@2.1.0:
+    resolution: {integrity: sha1-DGBSXqlHSPez85TcENbrjurE/jk=, tarball: https://ms-feed-25.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/ts-proto-descriptors/-/ts-proto-descriptors-2.1.0.tgz}
+
+  ts-proto@2.12.3:
+    resolution: {integrity: sha1-36Mqjfi7Xk5hKd/FLWaIgiNCQdc=, tarball: https://ms-feed-25.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/ts-proto/-/ts-proto-2.12.3.tgz}
```
mddinbox: The latest lockfile resolution embeds `ms-feed-25.pkgs.visualstudio.com` tarball URLs for `ts-proto` and its transitive dependencies (including `ts-proto` here, plus `case-anything`, `detect-libc`, `dprint-node`, `ts-poet`, `ts-proto-descriptors`, and the modified TypeScript entry). The current Lint job fails in `pnpm install` with `ERR_PNPM_LOCKFILE_RESOLUTION_VERIFICATION` because the active supply-chain policy compares these URLs with npm’s canonical `registry.npmjs.org` metadata. Please regenerate this lockfile from the canonical registry (or remove the private-feed tarball overrides) before merging; otherwise every PR check that installs dependencies is blocked.

mohamedmansour: Replaced the private-feed tarball overrides with the canonical npm integrity records for `ts-proto` and its transitive dependencies. A frozen `pnpm install --ignore-scripts` now passes the repository supply-chain policy, and the desktop generator suite remains green (28/28).

crates/webui-desktop-build/src/emit/typescript.rs
```diff
@@ -0,0 +1,280 @@
...
+            writeln!(out, "encode: () => {codec}.encode({{}}).finish(), decode: bytes => {{ {codec}.decode(bytes); }},\nvalidate: value => {{ if (value !== undefined) throw new IpcError('invalid-payload', 'Empty requires undefined'); }},").ok();
+        } else {
+            writeln!(out, "encode: value => {codec}.encode(value).finish(), decode: bytes => {codec}.decode(bytes),\nvalidate: (value, limits) => validateValue(value, {i}, messageShapes, limits),").ok();
+        }
+        writeln!(out, "validateBytes: (bytes, limits) => validateMessage(bytes, {i}, messageShapes, limits),\n}};").ok();
+    }
+    // `wireVersion` must match `webui_desktop::ipc::IPC_VERSION` (currently 3):
+    // the generated renderer and the framework's Rust host ship in the same
+    // binary and are never independently versioned.
+    writeln!(out, "export const schema: ConnectionSchema = {{ hello: {{ wireVersion: 3, contractName: {:?}, contractMajor: {}, schemaHash }}, methods: [", contract.name, contract.major).ok();
```
mddinbox, line 62: The generated renderer contract is being bumped to wire version 3 here, but the committed native bootstrap asset still contains `wireVersion!==2`. This leaves the checked-in generated asset out of sync and causes all three desktop release-consumer jobs to fail at `sync-bootstrap.mjs --check`. Regenerate `crates/webui-desktop/src/generated/ipc/native-bootstrap.js` with the explicit `--write` flow and commit the result.

mddinbox (changes requested):
The generator now emits `hello.wireVersion: 3`, but the checked-in `crates/webui-desktop/src/generated/ipc/native-bootstrap.js` still rejects every handshake whose version is not 2. The release-consumer acceptance step consequently fails on Linux, macOS, and Windows with `Generated IPC asset differs: native-bootstrap.js`; if shipped, the stale asset would reject the v3 handshake as `invalid-payload`. Please regenerate and commit the checked-in bootstrap with the repository's explicit `sync-bootstrap.mjs --write` flow, then rerun the release-consumer checks.

crates/webui-desktop-build/src/emit/typescript.rs
mohamedmansour: Fixed: I regenerated the checked-in native bootstrap asset so it matches the wireVersion=3 desktop contract, and the repo gate is passing. Thanks for catching the generated artifact drift.

mddinbox (dismissed):
Reviewed the latest head 46dd9cf, including the one-commit delta that regenerated the checked-in desktop runtime asset to match the v3 wire contract. The previously reported bootstrap/version mismatch is fixed, all prior authored threads are resolved, and the latest PR Checks run is green across the desktop build and native IPC validation. No remaining confirmed issues.

.github/workflows/pr.yml
```diff
@@ -60,6 +84,45 @@ jobs:
...
+      - name: Build workspace and examples
+        run: |
+          cargo build --workspace --exclude xtask --features microsoft-webui-desktop/cli,contact-book-desktop/source
+          cargo xtask build-examples
+
+      - name: Install native IPC test display
+        run: sudo apt-get install -y -qq apparmor dbus-x11 xvfb xauth
+
+      - name: Fetch native IPC fixture dependencies
+        run: cargo fetch --locked --manifest-path crates/webui-desktop/tests/fixtures/native-ipc/Cargo.toml
```
mddinbox, line 104: The newly added locked fetch fails on the current head in all three platform jobs (PR Checks run 35823332602): Cargo reports that it cannot update `crates/webui-desktop/tests/fixtures/native-ipc/Cargo.lock` because `--locked` was passed. This means the checked-in fixture lockfile does not match the manifest/resolved feature graph, so the job exits before the native IPC fixture runs and the PR cannot pass its merge gate. Regenerate the fixture lockfile from this manifest (including the native platform feature graph) and commit the resulting `Cargo.lock`, then keep `--locked` to preserve reproducible CI.

mddinbox (changes requested):
The newly added locked fetch fails on the current head in all three platform jobs (PR Checks run 35823332602): Cargo reports that it cannot update `crates/webui-desktop/tests/fixtures/native-ipc/Cargo.lock` because `--locked` was passed. This means the checked-in fixture lockfile does not match the manifest/resolved feature graph, so the job exits before the native IPC fixture runs and the PR cannot pass its merge gate. Regenerate the fixture lockfile from this manifest (including the native platform feature graph) and commit the resulting `Cargo.lock`, then keep `--locked` to preserve reproducible CI.

.github/workflows/pr.yml
mddinbox: Verified on the latest head `8d209f5a`: the fixture lockfile now matches the manifest—the stale `prost` dependency is removed—and the latest PR Checks run passes Build (Linux), Build (Windows), and Build (macOS). The locked-fetch failure is resolved. I’m leaving approval withheld because the same run still has an unrelated Ubuntu E2E failure in the pre-existing press fixture, so validation is not fully sufficient for approval.

examples/app/contact-book-manager/desktop/src/main.rs
```diff
@@ -0,0 +1,918 @@
...
+            2
+        );
+    }
+
+    #[cfg(feature = "source")]
+    #[test]
+    fn source_and_packaged_frames_render_app_owned_chrome() {
+        use webui_desktop::{build_desktop_bundle, DesktopBundleOptions, DesktopShellConfig};
+
+        let source = source_frame().unwrap();
```
mddinbox, line 708: This new test panics on a clean checkout before it can exercise the packaged frame. `source_frame()` unconditionally sets `config.asset_root` to `examples/app/contact-book-manager/dist`, but that directory is generated by the frontend build and is absent in the CI checkout; `DesktopApp::build()` canonicalizes the path and returns `NotFound`, which this `unwrap()` turns into the failing test seen in the current Test job. Please either make the source fixture create/use a checked-in or temporary asset root, or only configure `asset_root` when the directory exists so the test can run without a prior frontend build.

mohamedmansour: Fixed in 029603ca: both source-enabled tests now create isolated templates, state, assets, theme and a valid projection manifest, without making production asset or metadata checks optional. I reproduced the failures in a tracked-only checkout with no `dist` or frontend packages, confirmed all 12 app tests pass there after the fix, and ran the full quality gate successfully.

mddinbox (approved):

PR #368, opened 2026-06-24
fix: improve webui-press search results

mohamedmansour:
## Summary

Improve the built-in webui-press search experience and supporting docs shell behavior:

- Rebuild the search index extraction to decode escaped HTML text, preserve inline code in headings, and emit structured heading metadata with text, anchor, and level.
- Add weighted search ranking so page titles rank highest, matching section headings rank next, and body/path matches rank lower with deterministic tie-breaking.
- Render matching section headings as their own anchor-linked result rows, shown as page breadcrumbs such as `Page > Section`.
- Replace search result string interpolation with WebUI template state so titles/snippets render safely without `innerHTML`, including backtick code spans like `<for>`.
- Add safe match highlighting with theme-aware colors and a versioned two-phase render reset to prevent stale nested-loop DOM while typing quickly.
- Move docs page scrolling into `.main-content`, preserve sidebar/main scroll positions, support keyboard/page scrolling, and keep the browser viewport stable.
- Add darker themed scrollbars for sidebar/main content and a darker dark-mode search highlight.
- Add colocated TypeScript browser regression tests for docs-search behavior and wire all webui-press `*.test.ts` files into the docs gate via the new `@webui/webui-press` package script.

## Validation

- `cargo xtask check`

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -562,10 +562,12 @@ impl Step {
         run: || {
             // Build the standalone webui-press binary
             run_command_quiet("cargo", &["build", "-p", "microsoft-webui-press"], None)?;
-            // `@webui/webui-press test` builds the docs package first, then
-            // compiles and runs all colocated webui-press `*.test.ts` browser
-            // tests against that fresh output.
-            run_command_quiet("pnpm", &["--filter", "@webui/webui-press", "test"], None)
+            // `@webui/docs...` builds the docs package AND all of its workspace
+            // dependencies (e.g. `@microsoft/webui-framework`'s tsc compile)
+            // so that esbuild can resolve their `exports` to real .js files.
+            // Without this, a fresh checkout would fail with
+            // "Could not resolve @microsoft/webui-framework".
+            run_command_quiet("pnpm", &["--filter", "@webui/docs...", "build"], None)
         },
     };
     const BENCH_VALIDATE: Self = Self {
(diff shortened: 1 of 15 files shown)


xtask/src/main.rs
```diff
@@ -562,12 +562,10 @@ impl Step {
...
-            // `@webui/docs...` builds the docs package AND all of its workspace
-            // dependencies (e.g. `@microsoft/webui-framework`'s tsc compile)
-            // so that esbuild can resolve their `exports` to real .js files.
-            // Without this, a fresh checkout would fail with
-            // "Could not resolve @microsoft/webui-framework".
-            run_command_quiet("pnpm", &["--filter", "@webui/docs...", "build"], None)
+            // `@webui/webui-press test` builds the docs package first, then
+            // compiles and runs all colocated webui-press `*.test.ts` browser
+            // tests against that fresh output.
+            run_command_quiet("pnpm", &["--filter", "@webui/webui-press", "test"], None)
```
janechu: These new browser tests don't appear to run in PR validation yet: the Docs workflow still calls `pnpm --filter @webui/docs... build`, and the only workflow step that installs Chromium is the E2E job. Since this script launches Playwright `chromium`, please wire `@webui/webui-press test`/`cargo xtask docs` into the docs check with `pnpm exec playwright install --with-deps chromium` first. Otherwise the search/scroll regressions covered by these tests can merge without being exercised.

mohamedmansour: @Janechu I moved it to E2E, its natural home! thank you

janechu (approved):

PR #373, opened 2026-06-26
feat: add per-page script bundling to webui-press

mohamedmansour:
WebUI Press pages need page-scoped module scripts that can import npm packages without forcing every page to load the same global bundle. This adds an esbuild-based pipeline so markdown and custom pages can declare bundled scripts, while local component scripts are automatically included only where they are needed.

## Summary

- Extracts `<script type="module" bundle>` entries from page content and supports `customPages[].scriptFile` for custom pages.
- Discovers local component script pairs using the `tag-name.html` plus `tag-name.ts` convention and scans page/component HTML for local custom elements.
- Emits one shared root `index.js` for template chrome, and only emits page-specific code when content-specific component scripts or explicit bundled imports are needed.
- Reuses identical page import sets and collapses import-only `page-N.js` wrappers so pages link directly to the shared chunks they need.
- Keeps package custom-element registration explicit through page bundled scripts, for example `import "@mai-ui/button/define.js";`.
- Runs one esbuild build for root and page entries with ESM splitting, tree shaking, shared chunks, and build-mode minification.
- Moves script bundling, component discovery, graph validation, esbuild command construction, and injection helpers into `bundler.rs` so `build.rs` stays focused on orchestration.
- Hardens bundled script handling by sandboxing file paths and relative imports, rejecting absolute filesystem imports, skipping `node_modules` resolution when no scripts need bundling, using deterministic page signatures, and cleaning temp dirs through a guard.
- Documents the root/page script and bundler behavior in the WebUI Press README.

## Validation

- `cargo test -p microsoft-webui-press`
- `cd docs && cargo run -p microsoft-webui-press -- build --config .webui-press/config.json`
- `cargo xtask check`

## Performance check

A temporary docs page plus local component verified the end-to-end behavior with a local component auto-import and an explicit package import. The clean docs build now emits one root script, one real page bundle for the playground, and four shared chunks for local components used by markdown pages. Import-only page wrappers are removed, pages with no content-specific scripts load only `/webui/index.js`, and no non-WASM JavaScript files have duplicate content hashes.


crates/webui-press/src/build.rs
mcritzjam, line 1: step 2B happens after step 3 in linear order in this code.  which reads wierd.  not sure I love this step 1/2/2b/3/4/5/6/7/8/8b kind of structure.  lots to keep in your head.

mcritzjam (dismissed):

akroshg (approved):

PR #376, opened 2026-06-30
feat: add shared state to WebUI Press

mohamedmansour:
Add top-level state/stateFile loading for global render state and merge it into every generated page, including the 404 page.

Move shared and custom page state loading into a dedicated module with tests for merge precedence, reserved key protection, and state file validation.

This is needed so we can add dynamic data to press styled components with no custom pages.

### DIFF
--- a/crates/webui-press/src/state.rs
+++ b/crates/webui-press/src/state.rs
@@ -55,6 +55,7 @@ const RESERVED_STATE_KEYS: &[&str] = &[
     "next",
     "pageData",
     "headTags",
+    "tokens",
     "label",
     "icon",
 ];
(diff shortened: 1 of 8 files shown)


crates/webui-press/src/state.rs
```diff
@@ -0,0 +1,401 @@
...
+                    self.custom_pages.get(with_slash.as_str())
+                }
+            })
+    }
+}
+
+/// Top-level state keys reserved by the docs renderer. Global and custom-page
+/// state objects whose top-level fields are flattened onto the page state must
+/// not shadow these names so the canonical docs state always wins.
+const RESERVED_STATE_KEYS: &[&str] = &[
```
janechu, line 47: `tokens` should be treated as a reserved renderer-owned key here as well. Shared/custom state is merged before `inject_theme_tokens`, and the token injector writes CSS to `state.tokens` but silently does nothing if an existing `tokens` value is not an object. A config like `"state": { "tokens": "..." }` (or custom page state) would therefore prevent theme CSS from being injected and break templates that read `{{{tokens.light}}}` / `{{{tokens.dark}}}`. Please add `tokens` to the reserved list and keep the README reserved-key list in sync.

janechu (approved):

PR #378, opened 2026-07-02
feat: add plugin-owned FAST component discovery

janechu:
## Summary

- Adds a discovery plugin contract that maps framework-specific local and npm package layouts into WebUI component registrations.
- Adds FAST discovery for ordinary component files and CEM-backed `<component>.template.html` packages, shared by `fast_v2` and `fast_v3`.
- Converts supported `<f-template>`, `<f-when>`, and `<f-repeat>` sources for SSR while preserving FAST client templates, bindings, shadow options, and styles.
- Keeps validation, duplicate detection, CSS processing, render policy, caching, and runtime component output framework-neutral.

## Behavior

FAST discovery and source transformation run only with `fast_v2` or `fast_v3`, pinned to FAST major versions 2 and 3. The no-plugin and `webui` paths retain the native hyphenated-file and `template-webui.html` package layouts and treat FAST markup as ordinary authored HTML.

Malformed or unsupported FAST templates fail with located, actionable diagnostics. Discovery caches are isolated by plugin and invalidated by every package file that affects the selected layout.

## Validation

- `cargo xtask check`
- `todo-fast` Playwright suite
- `@fluentui/web-components@3.1.2`: all 42 FAST 3 templates with `link`, `style`, and `module` CSS


Cargo.toml
```diff
@@ -47,6 +47,7 @@ expand-tilde = "0.6.1"
 mime_guess = "2.0.5"
 html-escape = "0.2.14"
 include_dir = "0.7.4"
+microsoft-fast-convert = "0.9.0"
```
mohamedmansour: We don't want to add dependencies outside, we are trying to limit outside dependencies.

mohamedmansour (changes requested):
To be webui compliant, we want component files to be authored a specific way, neutral way, so that we can control the dom strategy, css strategy. We are doing a lot of things with that component discovery so making it consistent is what we need now.

If we want component discovery to have discover different formats like f-templates, or .webui files (like vue/astro), then that is a different feature request, that we need to plan.

But I don't see any value for us to complicate the plugin process for parsing a different schema for component discovery. For WebUI the schema must follow the Web Standard Syntax, intentionally. If you really want this, maybe have a pre-install script that transforms f-templates into webui schema.

Cargo.toml
janechu: OK, I can remove.

janechu:
> To be webui compliant, we want component files to be authored a specific way, neutral way, so that we can control the dom strategy, css strategy. We are doing a lot of things with that component discovery so making it consistent is what we need now.
> 
> If we want component discovery to have discover different formats like f-templates, or .webui files (like vue/astro), then that is a different feature request, that we need to plan.
> 
> But I don't see any value for us to complicate the plugin process for parsing a different schema for component discovery. For WebUI the schema must follow the Web Standard Syntax, intentionally. If you really want this, maybe have a pre-install script that transforms f-templates into webui schema.

OK, let me update to use a pre-install script.

My reasoning is that we can’t expect developers authoring FAST web components to also publish WebUI templates, as this would complicate testing. Similarly, if other web component libraries introduce their own declarative template syntaxes, aligning them with this integration pattern would be difficult. Instead, developers should be able to consistently use a single declarative template syntax or import templates from a published component library, avoiding this added complexity.

crates/webui-parser/src/plugin/fast_convert.rs
```diff
@@ -0,0 +1,297 @@
...
+    let (f_template_close, _) = find_matching_end(source, f_template.name, f_template_end)
+        .ok_or_else(|| {
+            ConvertError::new(
+                ConvertErrorKind::UnclosedElement {
+                    tag: f_template.name,
+                },
+                f_template_start,
+            )
+        })?;
+    let artifact = f_template_end..f_template_close;
```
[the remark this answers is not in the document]
janechu: Confirmed on latest head (43d6bf5d): the FAST `<f-template>` converter rejects any non-whitespace/non-comment content both before and after the sole wrapper (see `ContentOutsideTemplate` in `crates/webui-parser/src/plugin/fast_convert.rs`), returning a located `invalid-fast-template` diagnostic instead of silently dropping it. Regression tests cover both rejection directions as well as allowed whitespace/comments around the wrapper. Resolving this thread.

mohamedmansour (approved):

marchbox (approved):

chrisdholt (approved):

mcritzjam (approved):

PR #382, opened 2026-07-07
fix: warn when hydration leaves observable state disagreeing with SSR DOM (#379)

### LINKED ISSUE #379: [Bug]: Hydration leaves an element's DOM disagreeing with its own observable state (silent, and inconsistent with CSR)
### Before filing

- [x] This is not a security vulnerability.

### Area

WebUI Framework package

### Summary

After SSR hydration, an element's DOM keeps the server-rendered value while its observable holds a different client value — the two disagree permanently, with no warning. The identical component renders the value correctly when created client-side (CSR), so SSR and CSR diverge for the same code and the same value.

### Minimal reproduction

Two self-contained npm packages are attached (`if-hydration-repro/` for a structural `<if>`, `attr-hydration-repro/` for an attribute). No webui checkout or bundler — the framework loads via an `<script type="importmap">` and `webui serve` does the SSR. Each renders the same component **four times server-side and four times client-side** (value set in `constructor`, in `connectedCallback` before/after `super`, and deferred) and prints an SSR-vs-CSR table.

(A Playground link isn't possible — the bug needs SSR hydration, i.e. the CLI / native renderer.)

The component:
```html
<!-- app/repro-host/repro-host.html -->
<div class="container">
  <if condition="show"><div class="content">CONTENT</div></if>
</div>
```
```ts
// inline module script in app/index.html — plain JS, no decorators/bundler
class ReproHost extends WebUIElement {
  constructor() {
    super();
    if (this.getAttribute("when") === "constructor") this.show = true;
  }
  connectedCallback() {
    const when = this.getAttribute("when");
    if (when === "before-super") this.show = true;
    super.connectedCallback();
    if (when === "after-super") this.show = true;
    if (when === "defer") setTimeout(() => { this.show = true; }, 0);
  }
}
observable(ReproHost.prototype, "show");
```
The server renders with an empty state, so `show` is falsy server-side and the `<if>` is empty. (`attr-hydration-repro` is the same component, binding `data-value="{{value}}"` and setting `value = "3"`.)

### Steps to reproduce

[attr-hydration-repro.zip](https://github.com/user-attachments/files/29616489/attr-hydration-repro.zip)
[if-hydration-repro.zip](https://github.com/user-attachments/files/29616490/if-hydration-repro.zip)

1. `cd if-hydration-repro`   (or `attr-hydration-repro`)
2. `npm install`
3. `npm start`, then open http://localhost:3000
4. Observe the SSR-vs-CSR table on the page.

### Expected behavior

Either:
- the hydrated DOM reflects the observable's value, matching CSR of the same component (so an `@observable` default is honored under SSR too); or
- at minimum, a hydration-mismatch **warning** is emitted (dev builds) when an observable's value at `$ready` differs from what was server-rendered — as React / Vue / Svelte / Solid do.

Silently leaving an element's own state and DOM in disagreement is the core problem.

### Actual behavior

A value set on an observable **at or before `super.connectedCallback()`** (constructor, `@observable` class-field default, or before the `super` call) that differs from the SSR state is never reconciled. After hydration `host.show === true` but the `<if>` stays empty — no warning, no error.

- Bounded precisely by `super.connectedCallback()`: a value set *after* it (later in `connectedCallback`, a microtask / `setTimeout`, or a `@fromStore` subscription) reconciles fine.
- The same component created client-side (CSR / `createElement`) renders the value.
- Client-only values (`chrome.*` prefs, `localStorage`, feature detection) cannot be in the SSR state, so this isn't always avoidable by seeding state.
- Re-setting the same value later is a no-op (the observable didn't change), so an already-diverged value doesn't self-heal.
- Mirror of #286 (which trusts SSR content for *SSR-true → client-temporarily-false*); the reverse — a client value diverging from SSR at hydration — is never reconciled.

`if-hydration-repro` (`<if condition="show">`, `show` set to `true`):
```
timing         SSR       CSR
constructor    ABSENT    present   (diverges)
before-super   ABSENT    present   (diverges)
after-super    present   present
defer          present   present
```
`attr-hydration-repro` (`data-value="{{value}}"`, `value` set to `"3"`):
```
timing         SSR data-value    CSR data-value
constructor    ""                "3"   (diverges)
before-super   ""                "3"   (diverges)
after-super    "3"               "3"
defer          "3"               "3"
```

### Environment

- OS: Windows (the repros are OS-independent)
- Rust version: n/a (published CLI / native addon)
- Node.js version: 20+ (tested on 24)
- WebUI package or CLI version: `@microsoft/webui` + `@microsoft/webui-framework` **0.0.17**
- Command: `npm start` → `webui serve app --plugin webui --css style --servedir .`

### Logs or terminal output

```shell
No errors or warnings are produced — the failure is silent. The observed result is the SSR-vs-CSR tables above; the server render, the client bundle, and hydration all complete without any diagnostic.
```

mohamedmansour:
## Summary

Fixes #379. After SSR hydration, an element's public observable state could permanently and silently disagree with its own DOM. Any reactive write that runs while the element is *connected but not yet hydrated* — a `@observable` field initializer, the `constructor`, or code before `super.connectedCallback()` — updates the backing field but is dropped by `$update`'s pre-ready guard, which trusts the server-rendered DOM and leaves it untouched. Because the reactive setter no-ops when the same value is re-assigned later, the mismatch never self-heals: the element reports `show === true` while the DOM shows nothing, with no error and inconsistently with client-side rendering.

## Approach

The framework **intentionally does not reconcile** trusted SSR content. Patching the DOM after the fact would still leave the first server paint wrong and would erode the SSR-trust invariant that #286 relies on (a child may legitimately hydrate a complex `:prop` before its parent assigns it). Instead, this PR emits a **hydration-mismatch warning** — the same signal React, Vue, Svelte, and Solid produce — naming exactly which observables diverged and how to fix it: seed the value through SSR state, or assign it after `super.connectedCallback()`.

The check is **read-only**, **allocation-free on the correct path** (the tracking `Set` stays `null` for well-behaved components), and **precise**: it warns only when a recorded pre-ready write yields rendered output that actually differs from the DOM. It deliberately stays silent for render-invariant changes (e.g. `count` 5→3 under `count > 0`), unbound observables, complex `:prop` bindings (#286), and form-control `value`/`checked`/`selected`.

## What changed

- **Diagnostic runtime** (`template-element.ts`) — records connected pre-ready write paths and, once hydrated, compares each against the DOM, emitting one `console.warn` per component tag.
- **Extracted module** (`hydration-mismatch.ts`) — the pure SSR/DOM comparators live in their own file as free functions over a small resolver context, keeping the core class focused on lifecycle wiring and the rationale beside the code. Includes per-tag dedup so a mismatch shared by many instances reports once.
- **Tests** — a 6-component Playwright regression fixture (constructor / field-initializer / before-super warn; after-super / deferred / seeded stay silent) proves end-to-end behavior and fails on `main`; 25 unit tests directly assert the comparators' precision (render-invariant, `:prop`/form-control skips, template/boolean/plain attrs, text parts/raw, repeat length, dedup).
- **Docs** — `DESIGN.md`, the interactivity guide, `docs/ai.md`, and the framework README document the setup-timing rule.
- **Chore** — bumps `crossbeam-epoch` 0.9.18 → 0.9.20 (Cargo.lock only) to clear RUSTSEC-2026-0204, which was blocking the `cargo deny` gate.

## Behavior / contract

No change for correct components and **no DOM mutation** — SSR output stays authoritative. The only new behavior is a warning when an element's own observable state contradicts the server DOM after hydration.

## Validation

- Framework unit tests: **86 pass** (25 new)
- Full Playwright E2E: **132 passed / 2 pre-existing skips**; #286 complex-prop intact
- `cargo xtask check`: **all checks pass** (fmt, clippy, deny, test, build, wasm, examples, bench, docs)

Closes #379

### DIFF
--- a/packages/webui-framework/src/hydration-mismatch.ts
+++ b/packages/webui-framework/src/hydration-mismatch.ts
@@ -61,6 +61,30 @@ export interface PathBindings {
 }
 
 /**
+ * Entry point invoked by `TemplateElement` through a dynamic `import()`.
+ *
+ * Compares each recorded pre-ready write path against the SSR DOM and, if any
+ * diverged, emits the hydration-mismatch warning once. Routing the element
+ * through this function keeps every comparator and the message string in this
+ * module, reached only via that dynamic import — so a production bundler drops
+ * the whole module when `__WEBUI_DEV__` folds `DEV` to `false` (see
+ * `template-element.ts`). The comparison is read-only; it never mutates the DOM.
+ */
+export function reportHydrationMismatch(
+  tag: string,
+  writes: ReadonlySet<string>,
+  index: ReadonlyMap<string, PathBindings>,
+  ctx: MismatchContext,
+): void {
+  const mismatched: string[] = [];
+  for (const path of writes) {
+    const entry = index.get(path);
+    if (entry && bindingsDisagreeWithDom(entry, ctx)) mismatched.push(path);
+  }
+  if (mismatched.length !== 0) warnHydrationMismatch(tag, mismatched);
+}
+
+/**
  * True when any binding for a recorded pre-ready write disagrees with the DOM.
  * Stops at the first disagreement — the caller only needs to know whether the
  * observable diverged, not every binding that did.
--- a/packages/webui-framework/src/hydration-mismatch.ts
+++ b/packages/webui-framework/src/hydration-mismatch.ts
@@ -74,8 +74,13 @@ export function bindingsDisagreeWithDom(entry: PathBindings, ctx: MismatchContex
   return false;
 }
 
-/** Text binding: compare the resolved value against the rendered text/HTML. */
+/** Text binding: compare the resolved value against the rendered text. */
 export function textDiffersFromDom(b: TextBinding, ctx: MismatchContext): boolean {
+  // Raw `{{{expr}}}` bindings render via innerHTML, which the browser
+  // re-serializes on read (quote style, entity/attribute normalization, void
+  // elements), so a value that truly matches SSR can compare unequal. Skip
+  // them rather than emit a false mismatch.
+  if (b.raw) return false;
   let expected: string;
   if (b.parts) {
     expected = ctx.resolveParts(b.parts, b.scope);
@@ -85,8 +90,7 @@ export function textDiffersFromDom(b: TextBinding, ctx: MismatchContext): boolea
   } else {
     return false;
   }
-  const actual = b.raw && b.rawParent ? b.rawParent.innerHTML : b.node.data;
-  return actual !== expected;
+  return b.node.data !== expected;
 }
 
 /** Attribute binding: compare per kind, skipping cases SSR legitimately owns. */
@@ -120,7 +124,14 @@ export function condDiffersFromDom(c: CondBinding, ctx: MismatchContext): boolea
   return (c.instance != null) !== !!c.condition[0](ctx.resolver, c.scope);
 }
 
-/** Repeat binding: compare the collection length against the rendered items. */
+/**
+ * Repeat binding: compare the collection length against the rendered item
+ * count. Length is an intentional proxy — a same-length content or order change
+ * is not flagged, which keeps this cold-path check O(1) per repeat. Non-array
+ * values coerce to length 0, mirroring the renderer (`element/diff.ts` and the
+ * hydration path both treat non-arrays as empty), so the comparison stays
+ * consistent with what was actually rendered.
+ */
 export function repeatDiffersFromDom(r: RepeatBinding, ctx: MismatchContext): boolean {
   const resolved = ctx.resolveValue(r.collection, r.scope);
   const expectedLength = Array.isArray(resolved) ? resolved.length : 0;
@@ -140,15 +151,25 @@ export function formatHydrationMismatch(tag: string, paths: string[]): string {
 }
 
 /**
- * Tags already warned about in this document. A mismatch shared by many
- * instances of the same component (e.g. a list of 100 identical items) is one
- * authoring bug, so it reports once per tag instead of per instance.
+ * `(tag, path-set)` pairs already warned about in this document. Deduping on
+ * the exact set of diverged paths — not just the tag — keeps a repeated
+ * instance of the same bug quiet while still letting a genuinely different
+ * later mismatch on the same component surface.
  */
-const warnedMismatchTags = new Set<string>();
+const warnedMismatches = new Set<string>();
 
-/** Emit the hydration-mismatch warning once per tag. */
+/** Emit the hydration-mismatch warning once per `(tag, diverged-path-set)`. */
 export function warnHydrationMismatch(tag: string, paths: string[]): void {
-  if (warnedMismatchTags.has(tag)) return;
-  warnedMismatchTags.add(tag);
+  const key = `${tag}\u0000${[...paths].sort().join('\u0000')}`;
+  if (warnedMismatches.has(key)) return;
+  warnedMismatches.add(key);
   console.warn(formatHydrationMismatch(tag, paths));
+}
+
+/**
+ * Clear the dedup cache. Test-only: warnings dedupe for the lifetime of the
+ * document, so tests call this to assert warning behavior in isolation.
+ */
+export function resetHydrationMismatchWarnings(): void {
+  warnedMismatches.clear();
 }
--- a/packages/webui-framework/src/hydration-mismatch.ts
+++ b/packages/webui-framework/src/hydration-mismatch.ts
@@ -1,0 +1,154 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ * Hydration-mismatch diagnostic (issue #379).
+ *
+ * A reactive write that runs while an element is connected but before hydration
+ * finishes — a `@observable` field initializer, the `constructor`, or code
+ * before `super.connectedCallback()` — is dropped by `TemplateElement.$update`'s
+ * pre-ready guard: the backing field changes, but the server-rendered DOM is
+ * trusted and left untouched. That leaves the element's own observable
+ * disagreeing with its DOM — silently, and inconsistently with client-side
+ * rendering.
+ *
+ * The framework does NOT reconcile trusted SSR content: patching the DOM would
+ * only repair the post-hydration state while leaving the first server paint
+ * wrong, and it would erode the SSR-trust invariant (#286). Instead
+ * `TemplateElement` records the pre-ready write paths and, once hydrated, calls
+ * into this module to report any that disagree with the DOM — the same
+ * hydration-mismatch signal React, Vue, Svelte, and Solid emit.
+ *
+ * This module is intentionally free of element internals: the comparisons are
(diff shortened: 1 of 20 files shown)


mcritzjam (dismissed):
great test coverage

packages/webui-framework/src/hydration-mismatch.ts
```diff
@@ -0,0 +1,154 @@
...
+  let expected: string;
+  if (b.parts) {
+    expected = ctx.resolveParts(b.parts, b.scope);
+  } else if (b.path) {
+    const raw = ctx.resolveValue(b.path, b.scope);
+    expected = raw == null ? '' : String(raw);
+  } else {
+    return false;
+  }
+  const actual = b.raw && b.rawParent ? b.rawParent.innerHTML : b.node.data;
```
Qusic: Raw bindings compare `rawParent.innerHTML` against the raw string with `!==`. Browser `innerHTML` is re-serialized (quote style, entity encoding, `` -> `src="x"`, etc.), so a raw value that matches SSR but is not already in normalized form reports a false mismatch. The unit test uses a fake element whose `innerHTML` is a literal string, so normalization is never exercised. Consider normalizing both sides (or skipping raw bindings) before comparing.

packages/webui-framework/src/hydration-mismatch.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Conditional binding: compare rendered PRESENCE, not the raw value. A pre-ready
+ * write that changes the value but not the condition result (e.g. `count` 5→3
+ * under `count > 0`) leaves the DOM correct and must not warn.
+ */
+export function condDiffersFromDom(c: CondBinding, ctx: MismatchContext): boolean {
+  return (c.instance != null) !== !!c.condition[0](ctx.resolver, c.scope);
+}
+
+/** Repeat binding: compare the collection length against the rendered items. */
+export function repeatDiffersFromDom(r: RepeatBinding, ctx: MismatchContext): boolean {
```
Qusic, line 159: This only compares collection length, so a pre-ready write that changes item content/order but not length is missed. Also, a non-array iterable (Set/Map/generator) resolves to length 0 and would report a false mismatch if repeats accept those. Worth a comment noting length is an intentional proxy, or handling non-array iterables.

packages/webui-framework/src/hydration-mismatch.ts
```diff
@@ -0,0 +1,154 @@
...
+    `Add the value to the SSR state, or assign it after super.connectedCallback().`
+  );
+}
+
+/**
+ * Tags already warned about in this document. A mismatch shared by many
+ * instances of the same component (e.g. a list of 100 identical items) is one
+ * authoring bug, so it reports once per tag instead of per instance.
+ */
+const warnedMismatchTags = new Set<string>();
```
Qusic: `warnedMismatchTags` is module-global, per-tag, and never reset. A second, genuinely different mismatch on the same tag is suppressed for the whole session, and it is not document-scoped (persists across SPA renders). It is also not resettable, so the `warnHydrationMismatch` tests depend on unique tag names. Consider deduping per `(tag, paths)` and exporting a reset for tests.

packages/webui-framework/src/hydration-mismatch.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/** Emit the hydration-mismatch warning once per tag. */
+export function warnHydrationMismatch(tag: string, paths: string[]): void {
```
Qusic, line 186: `console.warn` and the message building are shipped and run in every build. React/Vue strip hydration-mismatch warnings from production. Since this only runs when pre-ready writes exist the cost is bounded, but consider a dev-only / tree-shakeable flag so the diagnostic path and message string do not ship to prod.

Qusic (commented):
A few non-blocking notes on the extracted diagnostic module. Nice refactor overall — the comparators being pure and unit-tested is a clear improvement.

One coverage note (not inline, since the anchor is outside the diff): `$applySSRState()` in template-element.ts (~line 376) overwrites the backing field for any observable whose key is in `__webui.state` before `$checkHydrationMismatch()` runs (~line 392), so the diagnostic can only fire for observables not covered by SSR state. That is fine, but the class comment reads as if it catches all pre-ready writes — might be worth narrowing the wording.

Qusic (dismissed):

packages/webui-framework/src/hydration-mismatch.ts
mohamedmansour: Good catch on the re-serialization — fixed in `3fe53b90`. `textDiffersFromDom` now skips raw `{{{expr}}}` bindings outright (`if (b.raw) return false;`) for exactly this reason: the browser re-serializes `innerHTML` on read (quote style, entity/attribute normalization, void elements), so a raw value that truly matches SSR can compare unequal. Skipping avoids the false positive without paying for normalization on this cold path. Covered by a unit test: `textDiffersFromDom › raw bindings are skipped (innerHTML re-serialization is unreliable)`.

packages/webui-framework/src/hydration-mismatch.ts
mohamedmansour: Addressed in `3fe53b90` with a rationale comment on `repeatDiffersFromDom`. Length is an intentional O(1) proxy — a same-length content/order change is deliberately not flagged on this cold path. On non-array iterables: there is no false mismatch, because the renderer itself treats non-arrays as empty (`element/diff.ts` and the hydration path), so a Set/Map/generator collection renders 0 items and `expectedLength` (0) matches `instances.length` (0). The comparison stays consistent with whatever was actually rendered.

packages/webui-framework/src/hydration-mismatch.ts
mohamedmansour: Reworked in `3fe53b90`. The dedup key is now `(tag, sorted path-set)` rather than tag alone (`warnedMismatches`), so a genuinely different later mismatch on the same component still surfaces while a repeat of the same bug stays quiet. Also exported `resetHydrationMismatchWarnings()` so the tests assert in isolation instead of leaning on unique tag names. The cache is intentionally document-lifetime — like React/Vue, a real hydration warning should not re-spam on every SPA navigation.

packages/webui-framework/src/hydration-mismatch.ts
mohamedmansour: Done in `ecbcef2f`. The diagnostic is now dev-only and fully tree-shakeable. `template-element.ts` reaches it only through a dynamic `import('./hydration-mismatch.js')` gated by `__WEBUI_DEV__` (`const DEV = typeof __WEBUI_DEV__ === 'undefined' || __WEBUI_DEV__`). When a bundler folds `__WEBUI_DEV__` to `false`, `$checkHydrationMismatch` empties and that `import()` is DCE'd, so the entire module — comparators, `formatHydrationMismatch`, and the `console.warn` — drops from the bundle (a static import would not strip, since esbuild fixes reachability before constant-folding). `webui-press build` injects `--define:__WEBUI_DEV__=false` automatically; `serve` leaves it on. It is on by default so raw ESM / tsc / unit tests keep the warning. Documented in the interactivity guide with a per-bundler define table.

Qusic (approved):

PR #397, opened 2026-07-18
Fix the helloworld.md

ccouzens:
Some of the examples use styles.css, so make sure it exists consistently.

state.json was inconsistently inside a data directory. Fix this.

We can assume the user is inside the directory, so don't specify hello-world on commands.

In rust, webui protocol is a separate crate

### DIFF
--- a/docs/tutorials/hello-world.md
+++ b/docs/tutorials/hello-world.md
@@ -11,6 +11,8 @@ hello-world/
 ├── src/
 │   └── templates/
 │       └── index.html
+├── assets/
+│   └── styles.css
 └── state.json
 ```
 
@@ -44,6 +46,13 @@ Create a simple template in `src/templates/index.html`:
 </html>
 ```
 
+## Creating the Styles
+
+Create a stylesheet in `assets/styles.css`. For focus, we'll keep it blank.
+
+```css
+```
+
 ## Creating the State
 
 Create a `state.json` file with the data for rendering:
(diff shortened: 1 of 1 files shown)


mddinbox (approved):
Reviewed the current head and the complete documentation diff. The project layout, serve/build commands, stylesheet asset, and Rust protocol import now match the current repository APIs. No issues found.

docs/tutorials/hello-world.md
```diff
@@ -77,30 +86,31 @@ This will:
...
 
 Here's how to render a pre-built protocol programmatically with Rust:
 
 ```rust
-use std::fs;
 use serde_json::Value;
-use webui_handler::{Protocol, RenderOptions, ResponseWriter, WebUIHandler};
+use std::fs;
+use webui_handler::{RenderOptions, ResponseWriter, WebUIHandler};
+use webui_protocol::WebUIProtocol;
```
mddinbox: This example now imports `webui_protocol::WebUIProtocol`, but the tutorial never adds `webui-protocol` as a direct Cargo dependency or tells readers to add it. In a normal Cargo application that follows the documented setup, this fails with an unresolved import because transitive dependencies are not available for direct use. Please either document the required `webui-protocol` dependency here or use the public `webui_handler::Protocol` API as before.

mddinbox (changes requested):
Read-only security/correctness review

Target: microsoft/webui#397
Base SHA: 4f0f45b523dacd8b44e4db4d2f11336fb60f66f8
Head SHA: da4517df2b5a04c117d0d8b31e8718a52ef126e3
Coverage: complete one-file diff, trusted-base contracts, existing review threads, and independent read-only challenge; CI checks were green.

Verdict: REQUEST_CHANGES

One introduced documentation correctness issue is noted inline. The updated Rust example imports webui_protocol directly, but the tutorial does not add or instruct users to add webui-protocol as a direct Cargo dependency.

PR #399, opened 2026-07-22
feat: add render state schema generation

Markionium:
## Summary

- Add `webui schema` to generate a render-state JSON Schema from an existing `protocol.bin`.
- Add `webui build --emit-schema` to emit `<protocol-stem>.state.schema.json` directly from the in-memory build result.
- Infer objects, arrays, scalar types, required and optional properties, component scopes, conditions, and integer predicates.
- Generate a separate schema definition for each complete route chain.
- Expose route-to-schema mappings through `x-webui-routes`.
- Add non-validating `x-webui.preferredType` metadata for downstream type generation.
- Add fixtures, snapshots, example-state validation, specifications, and CLI documentation.

## Why

The original proof of concept combined schema inference and C# generation in one spike. It also flattened route branches into one state shape, losing the distinction between the state required by individual routes.

This PR separates the language-neutral schema contract so it can be reviewed and evolved independently.

For non-routed applications, the output is a normal object schema.

For routed applications, the output contains one definition per complete route chain:

```json
{ 
  "$defs": {
    "route": {},
    "route.contacts.:id": {} 
  }, 
  "x-webui-routes": {
    "/": "#/$defs/route",
    "/contacts/:id": "#/$defs/route.contacts.:id"
  }
}
```

Route definitions remain independent even when two routes currently have the same state shape.

### Schema inference

The schema is inferred from how compiled templates use state:

• Plain rendered values and normal attributes accept scalar values.
• Raw values infer strings.
• Dotted paths infer nested objects.
•  `<for>` collections infer arrays and item shapes.
• Condition-only paths are optional because missing condition state evaluates false.
• Boolean conditions infer booleans.
• Ordered comparisons infer numbers.
• Integer literals infer  integer .
• Known types propagate across path-to-path equality.
• Component bindings resolve back to parent state paths.
• Complex properties preserve structured values.
• Internal handler signals are excluded.

Broad validation types can include a generation hint without narrowing the JSON Schema contract:

```json
{ 
  "type": ["string", "number", "boolean"], 
  "x-webui": {
    "preferredType": "string" 
  }
}
```

#### Why `x-webui.preferredType` exists

JSON Schema validation and DTO generation have slightly different goals.

The renderer is permissive. A normal text binding can render a string, number, or boolean:

```json
{ 
  "type": ["string", "number", "boolean"]
}
```

However, generated DTOs should guide developers toward the type normally intended by the template:

- Displayed values and normal attributes generally prefer string, because dates, counts, currency, and other values should usually be formatted before rendering.
- Condition flags prefer  boolean .
- Ordered numeric comparisons infer  number  or  integer  directly.
- Structured or genuinely unknown values remain unconstrained.

Languages such as C# do not have a native `string | number | boolean` union. Generating  object  would provide little type safety and complicate `System.Text.Json` source generation.

The schema therefore keeps the complete runtime-valid type while adding a non-validating annotation:

```json
{ 
  "type": ["string", "number", "boolean"], 
  "x-webui": {
    "preferredType": "string" 
  }
}
```

Standard JSON Schema validators use `type` and ignore `x-webui`. Host-language generators use `preferredType` to select a practical DTO type.

This keeps the schema accurate for dynamic consumers while producing useful strongly typed APIs for TypeScript, Rust, C#, and future generators.

#### Future: a WebUI JSON Schema vocabulary

`x-webui` is a pragmatic interim annotation. A better long-term design would be to publish a versioned WebUI render-state meta-schema and vocabulary built on JSON Schema Draft 2020-12.

That vocabulary could formally define:

- Route-to-state mappings.
- Preferred host-language types.
- WebUI-specific generation metadata.
- Allowed values and validation rules for those annotations.
- Versioning and compatibility guarantees for external tooling.

Generated schemas could then reference a WebUI-specific `$schema` URI rather than only the generic Draft 2020-12 meta-schema. Standard JSON Schema behavior would remain available, while WebUI-aware validators and generators could also validate the additional metadata.

This PR does not define that vocabulary. The current `x-webui` shape is intentionally provisional so the schema and generation workflow can be evaluated first.

### Validation

• Validate every repository example containing both  src/index.html  and  data/state.json .
• Snapshot component scope, nested collections, routed shells, and the complete Routes example.
• Cover nested and control-flow routes, optional and splat paths, nested arrays, complex properties,  .length , encoded route references, and literal-looking state keys.
•  cargo xtask check 
•  cd docs && pnpm build

### Status and limitations

This remains work in progress:

- Inference is based on template usage rather than an explicitly authored state declaration.
- Route support is substantially improved over the original prototype, but may still have uncovered route/template combinations.
- The schema is a build-time contract; runtime state is not currently validated against it.
- `x-webui` is a provisional annotation. A versioned WebUI JSON Schema vocabulary and meta-schema would be preferable long term.
- The implementation was largely AI-generated and still requires detailed human review before production use.

Host-language generation is intentionally excluded from this PR. A stacked follow-up demonstrates TypeScript, Rust, and C# generation on top of this schema contract.


crates/webui-cli/src/commands/state_schema/tests.rs
```diff
@@ -0,0 +1,547 @@
...
+        }
+    }
+    errors
+}
+
+fn schema_accepts_type(schema: &Value, actual: &str) -> bool {
+    match schema.get("type") {
+        Some(Value::String(expected)) => expected == actual,
+        Some(Value::Array(expected)) => expected.iter().any(|value| value.as_str() == Some(actual)),
+        _ => true,
```
mddinbox, line 460: This wildcard makes the validator skip `anyOf` schemas. For the `.length` union generated by `length_parent_schema`, any JSON type is accepted instead of only string, array, or object, so the example-state validation can pass invalid state. Evaluate `anyOf` alternatives here and add a regression case.

mddinbox (changes requested):
The schema test validator accepts every value when a schema has no top-level `type`, but `length_parent_schema` deliberately emits an `anyOf`-only schema for string/array/object lengths. As a result, `validate_state` never evaluates that union: a number, boolean, or null at a `.length` parent passes `example_state_files_match_generated_schemas`, so the test can report valid state for an invalid generated contract. Handle `anyOf` (and nested `properties`/`items`) before falling back to an unconstrained schema, and add a regression case with an invalid `.length` value.

PR #402, opened 2026-07-24
feat: return UTF-8 buffers from Node render

Qusic:
## Summary

- Make `Protocol.render()` return the canonical UTF-8 Node `Buffer`; remove the separate `renderBuffer()` API.
- Reuse the existing decoded protocol, handler/plugin, and rendering pipeline; `renderStream()` remains unchanged.
- Update Node callers, tests, examples, documentation, and benchmark semantics for the breaking return-type change.

Call `.toString("utf8")` explicitly when JavaScript string operations are required.

## Local benchmark

The old String and new Buffer implementations were measured in the same release run on Linux x64 with Node 24.18.0. Canonical `render()` now uses the measured Buffer path.

The Buffer path avoids creating a V8 JavaScript string and lets napi-rs expose the rendered Rust allocation as an external `Buffer`, so the benefit grows with output size.

| Contacts | HTML bytes | String P50 | Buffer P50 | Change |
|---:|---:|---:|---:|---:|
| 10 | 33,130 | 77.44 µs | 77.83 µs | ~0% |
| 100 | 245,464 | 866.78 µs | 770.04 µs | 11.2% faster |
| 1,000 | 2,370,550 | 9.08 ms | 7.20 ms | 20.7% faster |

Output bytes are identical across buffered and streaming paths.

## Validation

- `cargo xtask check`
- `cargo xtask bench node-addon --baseline before-opt`

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -142,7 +142,7 @@ V8/N-API string and callback crossings:
 
 - **Protocol construction** — Node `Buffer` to protobuf decode/index after the addon is loaded
 - **JSON-string render** — N-API conversion, JSON parse, Rust render,
-  and the returned JavaScript string
+  and the returned UTF-8 Node `Buffer`
 - **Object render** — the same path plus public-wrapper
   `JSON.stringify`
 - **Streaming first callback** — state conversion and JSON parse,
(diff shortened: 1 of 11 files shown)


mohamedmansour:
Thanks for measuring this. I’m fine with a breaking change here, and I’d prefer the API to have one canonical buffered render method rather than both `render()` and `renderBuffer()` differing only by return representation.

Please change `Protocol.render()` to return a UTF-8 `Buffer` and remove `Protocol.renderBuffer()`. The byte-oriented result is the correct default for SSR/HTTP output and is equal or faster in these benchmarks; callers that specifically need a JavaScript string can explicitly use `.toString("utf8")`. `renderStream()` should remain separate because it has different streaming semantics.

Please update the docs, examples, tests, and benchmark naming to reflect that API.

mohamedmansour (approved):

PR #406, opened 2026-07-27
fix: hydrate sibling repeats after nested repeats

### LINKED ISSUE #405: [Bug]: A sibling `<for>` after a root-level nested repeat hydrates against the wrong marker
### Before filing

- [x] This is not a security vulnerability.

### Area

WebUI Framework package

### Summary

During SSR hydration, a `<for>` following another `<for>` can bind to the preceding repeat’s nested `<!--wr-->` marker instead of its own marker.

This occurs when the first repeat contains a root-level nested repeat, so both nested and sibling repeat markers share the same DOM parent.

The second repeat’s SSR content remains visible but is not hydrated:its event handlers are not wired, and a later collection update inserts new nodes at the wrong nested anchor while leaving the original SSR nodes behind.

Reproduced against PR #403 at `49c6860`.

### Minimal reproduction

Component template:

```html
<div class="lists">
  <for each="group in groups">
    <span class="group-label">{{group.name}}</span>

    <for each="item in group.values">
      <span class="inner-value">{{item.value}}</span>
    </for>
  </for>

  <for each="other in others">
    <button class="other" @click="{select(other)}">
      {{other}}
    </button>
  </for>

  <output class="selected">{{selected}}</output>
</div>
```

Component class:

```ts
import { WebUIElement, observable } from "@microsoft/webui-framework";

export class TestRepeatSiblings extends WebUIElement {
  @observable groups = [];
  @observable others = [];
  @observable selected = "none";

  select(value: string): void {
    this.selected = value;
  }

  replaceOthers(): void {
    this.others = ["Three", "Four"];
  }
}

TestRepeatSiblings.define("test-repeat-siblings");
```

SSR state:

```json
{
  "groups": [
    {
      "name": "Color",
      "values": [
        { "value": "Black" },
        { "value": "Blue" }
      ]
    },
    {
      "name": "Size",
      "values": [
        { "value": "S" },
        { "value": "M" }
      ]
    }
  ],
  "others": ["One", "Two"],
  "selected": "none"
}
```

Regression test:

```ts
const host = page.locator("test-repeat-siblings");
const others = host.locator(".other");

await expect(others).toHaveCount(2);
await expect(others).toHaveText(["One", "Two"]);

await others.first().click();
await expect(host.locator(".selected")).toHaveText("One");

await page.evaluate(() => {
  (
    document.querySelector("test-repeat-siblings") as HTMLElement & {
      replaceOthers(): void;
    }
  ).replaceOthers();
});

await expect(others).toHaveCount(2);
await expect(others).toHaveText(["Three", "Four"]);
```

### Expected behavior

- The SSR-rendered `One` and `Two` buttons are hydrated.
- Clicking `One` updates `selected` to `One`.
- Replacing `others` removes/reuses the original second-repeat instances.
- Only `Three` and `Four` remain, in the second repeat’s original location.

### Actual behavior

- `One` and `Two` are visible from SSR, but clicking them does nothing.
- `selected` remains `none`, because the second repeat’s event bindings were never wired.
- Updating `others` inserts `Three` and `Four` at the first group’s nested-repeat anchor.
- The original SSR-rendered `One` and `Two` remain in the DOM.
- Nearby nested-repeat DOM can also be reordered or lose expected binding metadata.

The failure is silent and SSR-only.

### Root cause

Sibling repeat hydration advances with:

```ts
lastRepMarker = anchor;
```

The next repeat is then found using:

```ts
this.$findMarker(
  ssrParent,

  MARKER_REPEAT_START,
  lastRepMarker,
);
```

After the first repeat hydrates, a root-level nested repeat’s start `<!--wr-->` remains in the DOM as its runtime anchor. Searching after the outer start marker therefore finds that nested marker before the actual sibling marker.

`collectItemMarkers()` already returns the depth-matched outer `endMarker`, and end markers remain available until hydration resolution finishes.

### Suggested fix

Advance the sibling cursor past the matched repeat range:

```ts
const { items: itemMarkers, end: endMarker } =
  marker
    ? collectItemMarkers(anchor)
    : { items: [], end: null };

lastRepMarker = endMarker ?? anchor;
```

I applied this change locally and reran the regression above; the test passed, including initial event wiring and the subsequent collection replacement.

mohamedmansour:
A sibling `<for>` following a root-level nested repeat could hydrate against the nested repeat's start marker instead of its own, leaving SSR content visible but non-interactive and causing later updates to reconcile at the wrong anchor.

Advance the sibling repeat cursor to the depth-matched outer end marker returned by `collectItemMarkers()`. This skips the entire hydrated repeat range before locating the next sibling marker. Add an SSR fixture that verifies event wiring and collection replacement for the affected layout.

Closes #405

### DIFF
--- a/packages/webui-framework/src/template-element.ts
+++ b/packages/webui-framework/src/template-element.ts
@@ -1249,8 +1249,6 @@ export class TemplateElement extends HTMLElement {
           const insertRef = ssrParent.childNodes[Math.min(beforeIndex ?? staticCount, ssrParent.childNodes.length)] ?? null;
           ssrParent.insertBefore(anchor, insertRef);
         }
-        lastRepMarker = anchor;
-
         const repeatInsts: TemplateInstance[] = [];
         const hasCollectionState = this.$hasStateRoot(collection, scope);
         const itemsArr = this.$resolveValue(collection, scope);
@@ -1260,6 +1258,7 @@ export class TemplateElement extends HTMLElement {
         const { items: itemMarkers, end: endMarker } = marker
           ? collectItemMarkers(anchor)
           : { items: [] as Comment[], end: null as Comment | null };
+        lastRepMarker = endMarker ?? anchor;
 
         if (blockMeta && blockTplDom && anchor.parentNode && itemMarkers.length > 0) {
           if (
--- a/packages/webui-framework/tests/fixtures/nested-repeat/element.ts
+++ b/packages/webui-framework/tests/fixtures/nested-repeat/element.ts
@@ -111,5 +111,20 @@ export class TestNestedRepeatKeyedChain extends WebUIElement {
   }
 }
 
+export class TestRepeatSiblings extends WebUIElement {
+  @observable groups: NestedRepeatGroup[] = [];
+  @observable others: string[] = [];
+  @observable selected = 'none';
+
+  select(value: string): void {
+    this.selected = value;
+  }
+
+  replaceOthers(): void {
+    this.others = ['Three', 'Four'];
+  }
+}
+
 TestNestedRepeat.define('test-nested-repeat');
+TestRepeatSiblings.define('test-repeat-siblings');
 TestNestedRepeatKeyedChain.define('test-nested-repeat-keyed-chain');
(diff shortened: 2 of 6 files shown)


Qusic (approved):

mcritzjam (approved):
Looks fine to me.  I think you need some cleanup task to have a more expansive test coverage matrix - this is part of the text matrix you ought be able to generate and run.

PR #407, opened 2026-07-27
fix: make component registration idempotent for overlapping paths

akroshg:
The app directory and every `--components` path are scanned recursively. When a `--components` path overlaps the app directory (for example `--components ./src` while the app entry lives in `./src/foo`), the same component file gets discovered by both scans and the build failed with `Component '<tag>' is already registered`.

Registration is now idempotent: re-registering a tag with the same template, CSS, and client-ownership is a no-op. Re-registering a tag with different content is still reported as a conflict.

Covers all overlap directions (ancestor, descendant, equal) and duplicate `--components` entries. Adds registry unit tests and an end-to-end overlap build test; updates DESIGN.md and CLI docs.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1062,13 +1062,6 @@ SSR-only. Scriptless templates may contain server-rendered bindings,
 conditionals, and repeats, but they emit no client metadata or state.
 Local paths perform a recursive WalkDir scan for HTML files with hyphenated names, pairing matching CSS files — the same convention used by the parser's `ComponentRegistry`.
 
-Because the app directory is also scanned recursively, a `--components` path may
-**overlap** the app directory (for example `--components ./src` while the app
-lives in `./src/foo`). Registration is therefore **idempotent**: re-registering a
-tag with identical template content, CSS, and client-ownership — the same
-physical file discovered by both scans — is a silent no-op. Re-registering a tag
-with **different** content remains a hard `already registered` error.
-
 ### HTML Parser
 ```rust
 pub struct HtmlParser {
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1062,6 +1062,13 @@ SSR-only. Scriptless templates may contain server-rendered bindings,
 conditionals, and repeats, but they emit no client metadata or state.
 Local paths perform a recursive WalkDir scan for HTML files with hyphenated names, pairing matching CSS files — the same convention used by the parser's `ComponentRegistry`.
 
+Because the app directory is also scanned recursively, a `--components` path may
+**overlap** the app directory (for example `--components ./src` while the app
+lives in `./src/foo`). Registration is therefore **idempotent**: re-registering a
+tag with identical template content, CSS, and client-ownership — the same
+physical file discovered by both scans — is a silent no-op. Re-registering a tag
+with **different** content remains a hard `already registered` error.
+
 ### HTML Parser
 ```rust
 pub struct HtmlParser {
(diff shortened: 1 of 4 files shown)


DESIGN.md
```diff
@@ -1062,6 +1062,13 @@ SSR-only. Scriptless templates may contain server-rendered bindings,
 conditionals, and repeats, but they emit no client metadata or state.
 Local paths perform a recursive WalkDir scan for HTML files with hyphenated names, pairing matching CSS files — the same convention used by the parser's `ComponentRegistry`.
 
+Because the app directory is also scanned recursively, a `--components` path may
```
mohamedmansour: I don't believe we need this doc message, isn't it obvious that we are getting unique components

crates/webui-parser/src/component_registry.rs
```diff
@@ -125,6 +125,36 @@ impl ComponentRegistry {
...
+    /// re-registered with **different** content is a genuine conflict and
+    /// returns `ParserError::Component`.
+    ///
+    /// `css_definitions` and `css_fallback_chains` are derived deterministically
+    /// from `css_content`, so comparing `css_content` alone is sufficient.
+    fn resolve_duplicate(&self, candidate: &Component) -> Result<bool> {
+        match self.components.get(&candidate.tag_name) {
+            None => Ok(true),
+            Some(existing)
+                if existing.html_content == candidate.html_content
```
mohamedmansour: Why are we doing this, we just care about the component name for collisions, not content.

docs/guide/cli/index.md
```diff
@@ -577,6 +577,12 @@ webui build ./my-app --out ./dist --components ./shared/components
 webui build ./my-app --out ./dist --components /libs/ui-kit
 ```
 
+Because the app directory is scanned recursively too, a local `--components`
```
mohamedmansour: We don't need this doc message

crates/webui-parser/src/component_registry.rs
akroshg: Done — switched to name-based dedup. If the tag is already registered it's a no-op (first registration wins); no content comparison.

DESIGN.md
akroshg: Removed.

docs/guide/cli/index.md
akroshg: Removed.

mohamedmansour:
@akroshg with this PR we will now not know when components collide? Like if we add multiple "--components" since it is an array, of packages or locations, it could contain collisions and we want to know it is colliding. We should just remote the entry point component from --components wouldn't that work better

akroshg:
> @akroshg with this PR we will now not know when components collide? Like if we add multiple "--components" since it is an array, of packages or locations, it could contain collisions and we want to know it is colliding. We should just remote the entry point component from --components wouldn't that work better

In the general situation. We will have false positive as due to parent-child folder. eg. --components `./src`  and app file is at `./src/child`.
In this case this is not a collision, it is just we have revisited.
Should we store the whole path name so that during the discovery we can just see whether we have visited this path?
Do you see this as a common situation?

mohamedmansour:
After discussion, we’ve decided to keep the collision error because overlapping component registrations indicate a configuration issue that developers should correct.

A `--components` path must not overlap with the entry point’s component tree, and multiple `--components` paths must not collide with one another. We don’t currently support overlapping component roots, so these cases should fail explicitly rather than be made idempotent.

Closing this PR since the existing behavior is intentional.

PR #425, opened 2026-08-05
feat: add opt-in lazy hydration and rendering directives

mohamedmansour:
A long SSR list is costly at startup because work must be done every row: the browser styles, lays out, and paints all of them, and the framework upgrades and wires listeners on all of them, even when most rows sit far below the fold. This change introduces an opt-in, build-time policy that reduces that work to roughly what is on screen, without changing the server-rendered markup, the accessibility tree, or scroll geometry.

Eager remains the default. This is opt-in because the win only appears past a size threshold the engine cannot guess (see the 10-row row in the matrix below).

## Authoring surface

Two build-visible attributes on a component's root `<template>`:

```html
<!-- Recommended for rows repeated below the fold. -->
<template w-render="lazy" w-reserve-block-size="18rem">
  <article>...</article>
</template>

<!-- Advanced: defer only hydration, leave rendering untouched. -->
<template w-hydrate="lazy">
  <article>...</article>
</template>
```

Per-instance escape hatches:

```html
<product-card w-hydrate="eager"></product-card>
<product-card w-render="eager"></product-card>
```

`w-render="lazy"` composes `content-visibility: auto`, `contain-intrinsic-block-size: auto <reserve>`, and visibility-deferred hydration. `w-hydrate="lazy"` is the hydration half alone, for components where containment is unsafe for layout. Setting both is a build error.

`w-reserve-block-size` is required with `w-render="lazy"`. The compiler accepts one non-negative CSS length (absolute, font-relative, viewport, or container-query units) and rejects percentages, negatives, keywords like `auto`, and `calc()`.

## Why attributes instead of a TypeScript static

The previous revision used `static hydration = 'visible'`. That cannot work for the rendering half: the Rust compiler cannot see a TypeScript static, so `content-visibility` could not be applied until component JavaScript ran — which is after first layout, exactly when the cost has already been paid.

As build-time attributes, the compiler emits a deterministic, nonce-aware `<style data-webui-render-policy>` in the head, so containment applies before first layout. It also emits tree-scope-safe `:host(...)` and tag selectors so instances nested in another component's shadow root get the policy without it leaking to the parent.

The policy attributes are build-only: they are stripped, and the wrapper becomes the declarative shadow root for `--dom=shadow` or is unwrapped for `--dom=light`. Policy is carried in the protocol as a compact `wp: 1 | 2`.

## Correctness properties

The SSR subtree stays in the DOM. Nothing is removed or virtualized, so find-in-page, the full accessibility tree, focus order, selection, anchor links, and scroll anchoring keep working with no author code. The browser wakes a subtree via `contentvisibilityautostatechange`; pointer, focus, keyboard, and click activate a dormant instance synchronously ahead of observer delivery.

Also handled: parent-first activation for nested components, time-sliced hydration batches, streaming inserts, deferred state replay, reconnect after teardown, and eager fallback when `IntersectionObserver` is unavailable.

## Benchmark evidence

Chromium via Playwright 1.61.1 (headless), 1280x720. Medians of **30 timing runs**; render slices from **5 trace runs** over a window closing before hydration starts, so style/layout/paint are attributable to the browser rather than to script. Times in ms.

| n | mode | parse | style | layout | prePaint | paint | raster | render Σ | hydrate CPU | total | layout objs | DOM nodes | listeners | retained heap KiB |
|---:|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 10 | eager | 0.1 | 0.2 | 8.1 | 0.1 | 0.4 | 4.0 | 13.0 | 1.5 | **14.5** | 150 | 152 | 20 | 388.0 |
| 10 | lazy-hydrate | 0.1 | 0.2 | 7.9 | 0.1 | 0.4 | 3.6 | 12.2 | 1.9 | **14.1** | 150 | 152 | 20 | 434.6 |
| 10 | lazy-render | 0.1 | 0.3 | 8.4 | 0.2 | 0.4 | 4.0 | 13.4 | 1.9 | **15.3** | 160 | 152 | 20 | 435.4 |
| 100 | eager | 0.4 | 0.7 | 10.0 | 0.5 | 0.7 | 5.8 | 18.2 | 2.8 | **21.0** | 1500 | 1502 | 200 | 507.2 |
| 100 | lazy-hydrate | 0.4 | 0.7 | 9.4 | 0.6 | 0.7 | 4.8 | 16.6 | 2.6 | **19.2** | 1500 | 1502 | 26 | 491.7 |
| 100 | lazy-render | 0.4 | 0.5 | 7.8 | 0.3 | 0.4 | 4.9 | 14.3 | 2.6 | **16.9** | 475 | 1502 | 26 | 494.5 |
| 1000 | eager | 3.3 | 5.9 | 27.0 | 6.5 | 1.0 | 7.4 | 51.0 | 12.4 | **63.4** | 15000 | 15002 | 2000 | 1480.5 |
| 1000 | lazy-hydrate | 3.3 | 5.7 | 27.2 | 6.4 | 0.9 | 4.6 | 48.0 | 7.2 | **55.2** | 15000 | 15002 | 26 | 811.3 |
| 1000 | lazy-render | 3.3 | 1.7 | 8.5 | 0.9 | 0.4 | 5.5 | 20.4 | 8.0 | **28.4** | 1375 | 15002 | 26 | 828.9 |

At 1,000 rows, `w-render="lazy"` vs eager:

| Metric | eager | lazy-render | Δ |
|---|---:|---:|---:|
| Style recalc | 5.9 | 1.7 | **−71%** |
| Layout | 27.0 | 8.5 | **−69%** |
| Pre-paint | 6.5 | 0.9 | **−86%** |
| Layout objects | 15,000 | 1,375 | **−91%** |
| Forced style+layout to first render | 33.4 | 1.9 | **−94%** |
| Event listeners | 2,000 | 26 | **−99%** |
| Retained JS heap | 1,480.5 KiB | 828.9 KiB | **−44%** |
| **Total CPU** | **63.4** | **28.4** | **−55%** |

Three observations worth calling out:

1. **The two levels are nested, not independent.** `w-hydrate="lazy"` moves script cost only — layout is unchanged (27.2 vs 27.0 ms). `w-render="lazy"` defers style/layout/paint *on top of* that, and already includes lazy hydration: its listener count (26) and retained heap (828.9 KiB) match the hydrate-only arm rather than eager (2,000 / 1,480.5 KiB). The parser enforces the nesting — declaring both attributes is a build error (`INVALID_COMPONENT_RENDER_POLICY`: ``remove `w-hydrate`; `w-render="lazy"` already includes lazy hydration``), and the policy is a single three-state ladder (`Eager` -> `LazyHydration` -> `LazyRender`). Both spellings exist because the cheaper level is worth reaching for on its own: `w-render="lazy"` additionally requires `w-reserve-block-size` and is a net regression below ~100 rows.
2. **There is a hard floor.** Parse (3.3 ms) and DOM node count (15,002) are identical in all three arms. Containment defers work on nodes; it never avoids creating them. Anything below that floor needs a different primitive (declarative data-driven population), which is out of scope here.
3. **Scroll geometry is preserved.** Document height was 72,000 px in all three arms at n=1000 — the reason `w-reserve-block-size` is mandatory.

**Cost at small n:** at 10 rows `lazy-render` is a net regression — 15.3 ms vs 14.5 ms total (+6%) and +12% retained heap — because observer setup and the extra style pass exceed what there is to save. The crossover sits between 100 and 1,000 rows. This is the evidence for keeping it opt-in.

**Bundle:** eager is 14,016 gzip bytes; opting in produces 15,462 gzip bytes, a cost of 1,446 bytes. A test asserts eager-only module graphs exclude the coordinator, so projects that never opt in pay nothing.

Reproduce:

```bash
cd examples/integration/streaming-browser-bench
WEBUI_LAZY_HYDRATION_MODES=eager,lazy-hydrate,lazy-render \
WEBUI_LAZY_HYDRATION_RUNS=30 \
WEBUI_LAZY_HYDRATION_TRACE_RUNS=5 \
pnpm test:lazy-hydration
```

## Naming

Both values are `lazy`, not `visible`/`offscreen`, following review feedback from @gregwhitworth.

`visible` overclaimed. Four shipped behaviors hydrate content that is not on screen: the 200px lead margin, interaction wake while offscreen, native content-visibility relevance (find-in-page, focus, selection), and the eager fallback when `IntersectionObserver` is missing. Each has a test. `content-visibility: auto` likewise triggers on relevance rather than screen position, so `offscreen` had the same flaw.

`lazy`/`eager` also mirrors the platform's existing `loading="lazy"|"eager"` pair, replacing an asymmetric trigger-vs-timing pair, and leaves room for the definition to evolve.

## Validation

- `cargo xtask check` — all phases green
- `cargo test -p microsoft-webui-parser` — 544 pass
- `pnpm test:unit` (framework) — 227 pass
- `pnpm test:e2e` — 199 pass, 2 skipped
- `pnpm --dir docs build`

Docs: new Lazy Component Policy directive page (`docs/guide/concepts/directives/lazy.md`), plus updates to hydration, interactivity, performance, `DESIGN.md`, and both READMEs.


packages/webui-framework/src/element.ts
```diff
@@ -34,12 +47,145 @@ import type {
...
+  ): void {
+    // Call virtually so an authored override that instruments or extends the
+    // protected activation hook retains its existing behavior.
+    this.$activateDeferredSSR(state);
+  }
+
+  override disconnectedCallback(): void {
+    disconnectLazyHydration(this);
+    super.disconnectedCallback();
+  }
```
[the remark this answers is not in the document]
mohamedmansour: Added two fast paths in `6c7a53ec`: eager classes now skip `disconnectLazyHydration()` entirely, and the coordinator returns before generation or initial-observation bookkeeping for untracked visible instances. A regression test confirms eager disconnects never reach the coordinator.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

gregwhitworth:
This makes a ton of sense. Might I recommend us exploring expanding [loading](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attributes) to an element's attached event listeners. You're actually implementing this already in a similar fashion to how the standard recommends to derive the lazy algo. So minor recommendation would be to not have "visible" but adjust to `lazy` and this allows you to change the definition of lazy whereas visible has a pretty concrete definition. Additionally, within Open UI we've had requests for a declarative solution for lists within select or elsewhere that needs to do an initial population from data and possibly re-hydrate with lookups upon reaching a scrolling threshold when we finally introduce a searching functionality. I say all this to say that having this will be valuable to prove out the value add here and what declarative paths make the most sense. It being an attribute so that it can scale to other elements is the best approach. 

That's my 2 cents but I'm interested in other people's POV.

packages/webui-framework/src/lifecycle.ts
```diff
@@ -44,6 +45,33 @@ let started = false;
...
+let documentReady = !hasBrowserDocument ||
+  document.readyState === 'complete' ||
+  (navigationTiming?.domContentLoadedEventStart ?? 0) > 0;
+
+if (!documentReady) {
+  document.addEventListener('DOMContentLoaded', () => {
+    documentReady = true;
+    tryComplete();
+  }, { once: true });
+}
```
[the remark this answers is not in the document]
mohamedmansour: Fixed in 86a34b74 by adding a one-shot `load` fallback when DOMContentLoaded status cannot be established from navigation timing. This preserves the parser/module-startup gate instead of treating every interactive document as ready, while ensuring a late-loaded runtime cannot wait forever; the unit test covers both paths.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

packages/webui-framework/src/visible-hydration-coordinator.ts
```diff
@@ -0,0 +1,412 @@
...
+  ) return;
+  initialObservationTargets.add(target);
+  initialObservationCount++;
+  if (startupObservationGateActive) return;
+  startupObservationGateActive = true;
+  hydrationStart();
+  document.addEventListener('DOMContentLoaded', sealStartupObservationGate, {
+    once: true,
+  });
+}
```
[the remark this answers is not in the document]
mohamedmansour: Fixed in d43dc9bf by sealing the coordinator startup gate on `load` as well as `DOMContentLoaded`, matching the lifecycle fallback without weakening parser/module startup ordering. The regression test uses an interactive document with unavailable navigation timing and verifies the coordinator balances completion when `load` fires.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

packages/webui-framework/src/lazy-hydration-coordinator.ts
```diff
@@ -0,0 +1,413 @@
...
+function observeStreamed(
+  target: LazyHydrationTarget,
+  state: Record<string, unknown> | undefined,
+): void {
+  boundaryStates.set(target, state);
+  observe(target);
+}
+
+/** Stop observing a disconnected component without losing reconnect eligibility. */
+function disconnect(target: LazyHydrationTarget): void {
```
Qusic, line 114: `disconnect()` clears `observedTargets` and unobserves, but leaves the `boundaryStates` entry in place. A streamed target that is still dormant when it disconnects and reconnects re-registers through the ordinary `observe()` path (no state), yet `activate()` at L322 still sees `boundaryStates.has(target)` and replays the pre-disconnect boundary state through `$activateDeferredSSRFromBoundary()`.

The path is reachable: `invokeActivationHook` removes `STREAMED_HOST_ATTR` on the success path (streaming-deferred.ts L449) even when the hook only handed the element to the visibility coordinator, so on reconnect `$didDeferSSRHydration()` in element.ts takes the `!hasAttribute(STREAMED_HOST_ATTR)` branch and calls `observeLazyHydration(this)`. Triggered by any dormant streamed component that is recycled or moved: virtualized rows, a `<for>` reorder, or a DOM move.

Repro against this branch (`node --test`, same fake-global pattern as `visible-hydration-coordinator.test.ts`):

```ts
contract.observeStreamedLazyHydration(target, { from: "boundary-commit-1" });
contract.disconnectLazyHydration(target);
contract.observeLazyHydration(target);            // ordinary path, no state
ioCallback([{ target, isIntersecting: true, intersectionRatio: 1, time: 0 }]);
// activate() receives: { state: { from: "boundary-commit-1" }, streamed: true }
// expected:            { state: undefined,                     streamed: false }
```

Would clearing `boundaryStates` in `disconnect()` be the right fix, or should a reconnected dormant streamed host be re-registered through `observeStreamed()` so the retained state is intentional rather than incidental?

packages/webui-framework/src/lazy-hydration-coordinator.ts
```diff
@@ -0,0 +1,413 @@
...
+
+/** Whether this browser can defer hydration without risking an inert component. */
+function supportsVisibleHydration(): boolean {
+  return typeof IntersectionObserver !== 'undefined';
+}
+
+/** Register an ordinary SSR component for viewport and interaction activation. */
+function observe(target: LazyHydrationTarget): void {
+  pendingTargets.add(target);
+  if (!target.isConnected || observedTargets.has(target)) return;
```
Qusic, line 81: `pendingTargets.add(target)` runs before the `isConnected` check, so calling `observe()` on a disconnected target leaves it marked pending while skipping `observedTargets.add`, `advanceActivationGeneration`, `getObserver().observe()`, and `installWakeListeners()`.

That target is then unreachable by both routes: `enqueueActivation` returns early because `activationGenerations.get(target)` is `undefined`, and there is no observer entry to deliver an intersection. If no other target is currently observed, the wake listeners are not installed either, so interaction cannot reach it. The component stays dormant permanently.

Verified on this branch with a disconnected target:

```
IntersectionObserver.observe calls : 0
wake listeners on document         : []   (no pointerover/pointerdown/focus/keydown/click)
```

Moving `pendingTargets.add(target)` below the early return would make the state unrepresentable. Is there a case that relies on marking a disconnected target pending?

packages/webui-framework/src/lazy-hydration-coordinator.ts
```diff
@@ -0,0 +1,413 @@
...
+  initialObservationCount--;
+  finishStartupObservationGate();
+}
+
+function sealStartupObservationGate(): void {
+  startupObservationGateSealed = true;
+  finishStartupObservationGate();
+}
+
+function finishStartupObservationGate(): void {
```
Qusic, line 419: The startup gate requires `initialObservationCount === 0`, and that only decrements when an intersection entry is delivered (or on activate/disconnect). `sealStartupObservationGate` covers the DOMContentLoaded/load side, but there is no timeout on the count, so a single undelivered observer callback holds the `hydrationStart()` reference open permanently.

Verified on this branch: with one connected, observed target whose intersection callback never arrives, `webui:hydration-complete` never fires even after DOMContentLoaded and load have both been dispatched.

```
DCL + load dispatched, waited 300ms -> webui:hydration-complete count: 0
```

Since intersection callbacks are delivered as part of rendering updates, what is the intended behavior when those are throttled or suspended - a prerendered document, or a page loaded in a background tab? I could not reproduce either state under Playwright automation, so this is a question rather than a claim, but the absence of a timeout means the failure mode is permanent rather than delayed. Would a deadline that force-settles the remaining count (and reports it) be reasonable?

packages/webui-framework/src/lazy-hydration-coordinator.ts
mohamedmansour: Retaining boundaryStates is intentional: disconnect suspends observation but does not cancel deferred hydration; clearing it would lose the component's boundary-local bootstrap and fall back to unrelated global state on reconnect. Updates made before activation are recorded as deferred writes and win over the retained snapshot, so the same target must remain a streamed activation until it hydrates.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

packages/webui-framework/src/lazy-hydration-coordinator.ts
mohamedmansour: The disconnected state is intentional rather than permanently unreachable UI: ordinary registration comes from connectedCallback, while a detached streamed root may register before first connection; its later connectedCallback calls observe again and completes observer/listener setup. A node that remains disconnected cannot intersect or receive interaction, so moving the WeakSet write would not fix a reachable lifecycle failure.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

packages/webui-framework/src/lazy-hydration-coordinator.ts
mohamedmansour: A deadline would violate the IntersectionObserver contract we rely on: the specification tracks pending initial targets and prevents a fully active document from skipping the rendering update that queues their initial entries. Throttling may delay completion until a background or prerendered page resumes, but force-settling could declare startup complete while a visible root is still dormant, and the timer itself offers no stronger scheduling guarantee.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

Qusic (dismissed):

mcritzjam (approved):

PR #429, opened 2026-08-06
feat: make Light DOM a global CSS opt-in

### LINKED ISSUE #410: [Feature]: --dom light is unusable - no CSS delivery path and no <slot> projection
### Before filing

### Area

WebUI Framework package

### Problem or need

We ship a moderately large WebUI app (32 components, 41 stylesheets) as a **component asset** loaded from a CDN into host pages we do not control. Some of those hosts render the component inside a shadow root they own, so we tried moving our components to `--dom light` and letting the host provide the single encapsulation boundary.

`--dom light` compiles and emits an asset, but the result is not usable. Two independent gaps mean a light-DOM component asset cannot render correctly today, so `--dom light` is effectively shadow-only in practice.

**1. No CSS delivery path works in light mode.**

Same project, same source, only `--css` varying (compiler + framework `0.0.19`):

| `--dom light --css …` | Result |
| --- | --- |
| `link` | No `<link>` is emitted inside any template, and no aggregate stylesheet is emitted → nothing ever loads the CSS. |
| `module` | `templateStyles` **is** emitted (the `data:text/css` importmap, ~96 KB) and `registerAssetStyles` injects it into `document.head`, but `sa` is `undefined` on **all 32** templates → `injectModuleStyle` is never called, so the registered specifiers are never imported. |
| `style` | CSS is delivered, but inlined into every template, so it is duplicated per *instance* rather than per component, and the stylesheet text becomes part of the element's `textContent`. |

The `module` case looks like the intended light-DOM path, since `injectModuleStyle` already has the light-DOM branch:

```js
// element/styles.js
export function injectModuleStyle(specifier, shadowRoot) {
    if (shadowRoot) { /* adoptedStyleSheets on the shadow root */ }
    else if (!headInjected.has(specifier)) {
        headInjected.add(specifier);
        import(specifier, { with: { type: 'css' } }).then((mod) => {
            document.adoptedStyleSheets.push(mod.default);   // <-- light DOM path
        }, () => { });
    }
}
```

That branch dedupes by specifier and pushes to `document.adoptedStyleSheets`, which is exactly right for light DOM — it just never runs, because the compiler does not emit `sa` when `--dom light`.

Comparing the two builds directly:

```
shadow + module : sd=32  sa=32  keys=["h","c","sa","sd","tr","ta","b"]
light  + module : sd=0   sa=0   keys=["h","c","tr","ta","b"]
```

**2. `<slot>` is emitted verbatim but never projected.**

The compiler passes `<slot>` through unchanged in light mode. Four of our primitives project children, and all four are affected:

```
button    | shadow slots: 1 | light slots: 1
tab       | shadow slots: 1 | light slots: 1
tablist   | shadow slots: 1 | light slots: 1
divider   | shadow slots: 1 | light slots: 1
```

Outside a shadow root `<slot>` does not project, and the framework has no projection emulation — the only slot-related code in `template-element.js` is template *binding* slots for `<if>`/repeat blocks, which is a different concept. So authored children are dropped.

**Observed result.** Building our real app with `--dom light --css style` (the only mode that delivers CSS at all) and loading it in a browser:

```
rootHasShadow : false                 <- light DOM active, as expected
innerBg       : rgb(255, 255, 255)    <- class-based rules DO apply
rootDisplay   : "inline"              <- :host { display: block } is dead
tabLabels     : []                    <- no tabs rendered at all
buttonLabels  : [""]                  <- slot projection dead
headerText    : ":host { display: block; } .header { box-sizing: bo…"
```

The app renders as loading skeletons only — no content, no tabs, no images — with zero page errors, so nothing signals that the mode is unsupported.

Related and lower priority: compiled CSS is **byte-identical** between `--dom shadow` and `--dom light`. `:host` rules are left as-is (dead selectors in light DOM), and class selectors are not scoped, so every component's classes land in one global namespace. In our app that collapses 38 class names shared across components — `.panel` and `.panel-title` are each defined by 15 different components with different rules.

### Who would benefit?

Anyone shipping WebUI **component assets** into a host page they do not control. The host often already owns a shadow root, or needs to apply its own theming and layout to the embedded component, and does not want a second encapsulation boundary per component.

It also matters for hosts that cannot server-render — where there is no SSR endpoint available and the component must be fetched from a CDN and rendered entirely client-side. Light DOM is the natural fit there.

### Desired outcome

1. `--dom light` emits `sa` per template just as `--dom shadow` does, so `injectModuleStyle`'s existing `document.adoptedStyleSheets` branch runs and each component's CSS is registered exactly once per document.
2. `<slot>` is projected in light DOM — either the compiler lowers `<slot>` into an explicit insertion point the runtime fills from the element's authored children, or the framework emulates projection at hydration/wire time.
3. Ideally, `:host` is lowered to the component's tag selector when `--dom light`, so `:host`, `:host([attr])`, and `:host([attr])::before` keep working. Optional but very helpful: scope remaining selectors by tag to avoid the global-namespace collapse.
4. Failing all of the above, `--dom light` should **fail the build** (or warn loudly) when a component uses `<slot>` or `:host`, rather than emitting an asset that silently renders blank.

### Concrete example

```bash
webui build ./src \
  --entry index.html \
  --out ./dist \
  --plugin webui \
  --css module \
  --dom light \
  --emit-component-assets my-card
```

Given a component that projects children and styles its host:

```html
<!-- my-button.html -->
<button class="control" type="button"><slot></slot></button>
```

```css
/* my-button.css */
:host { display: inline-flex; }
.control { padding: 4px 12px; }
```

Used as `<my-button>Save</my-button>`, we would expect in light DOM:

- `sa: "my-button"` on the template, so `my-button.css` is imported once into `document.adoptedStyleSheets`
- `Save` projected into `.control`
- `:host { display: inline-flex }` lowered to `my-button { display: inline-flex }`

Today the label is dropped, the CSS is never imported, and the `:host` rule is inert.

### Constraints

- **Migration:** `--dom shadow` remains the default; this only needs to change behavior under `--dom light`, so existing shadow consumers are unaffected.
- **Compatibility:** `document.adoptedStyleSheets` and CSS module imports (`with { type: 'css' }`) are already relied on by the shadow path, so light DOM adds no new platform requirement.
- **Global CSS is inherent to light DOM**, so tag-scoping (item 3) is what keeps it tractable at our scale (32 components, 41 stylesheets). Without it, adopting light DOM means hand-editing every stylesheet.
- **Correctness over silence:** the current failure mode is the worst kind — a clean build, zero runtime errors, and a blank component. Even just item 4 would have saved us the investigation.

### Alternatives or workarounds

- **Staying on `--dom shadow` and letting the host mount the component inside its own shadow root.** This is what we shipped. Nested shadow roots work fine, custom properties pierce the boundary, and we verified the full app renders and is interactive inside a host-owned shadow root. It is a good outcome, but it does not give the host the ability to style component internals.
- **A mixed asset** — light for our own components (none of which use `<slot>`), shadow for the primitives that do. The runtime already supports this, since `sd` is a per-template flag and `wantShadow = hasShadow || !!meta.sd`, so merging a light build with the shadow build's entries for the slot-using primitives produces a valid asset. But the compiler only exposes `--dom` globally, so this requires post-processing the emitted asset, plus hand-rewriting `:host` and resolving the class collisions. A per-component `--dom` override, or a template-level opt-in, would make this a supported path.
- **Wrapping templates in `<template shadowrootmode="open">`** to opt individual components back into shadow under `--dom light`. This does not work: the compiler strips the wrapper and keeps only its contents (`shadowrootmode` occurrences in the emitted asset: 0).

### LINKED ISSUE #433: [Feature]: `--css-bundle` with shared chunk splitting, reusing the component-asset chunker
### Before filing

### Area

Performance

### Problem or need

CSS delivery emits **one stylesheet per component**, with no grouping step. A single route ships one render-blocking `<link>` per component under `Link` (`contact-book-manager`: 7 on one page), or one `<style>` block per component under `Module` (`commerce`: 15).

This costs on two independent axes, both measured.

**1. Compression.** Per-component files each carry their own compression window, so the shared token vocabulary (property names, marker attributes, color literals) can't be exploited across them:

```
contact-book-manager, 7 component CSS files
  7 separate files : raw 19,240 B   gzip 5,049 B
  1 bundled file   : raw 19,241 B   gzip 3,122 B
  saving           :                gzip 1,927 B  (38.2%)
```

**2. Style recalculation.** The engine walks each stylesheet's rule buckets per element, so sheet *count* carries a real per-element cost independent of rule count. Measured by holding the rules byte-identical and varying only how many `CSSStyleSheet` objects they're spread across. All arms verified to produce identical computed styles across ~16,500 values before timings were trusted:

| page | split (as shipped) | bundled | atomized (1 rule/sheet) |
|---|---|---|---|
| product | 15 sheets · 1.271 ms | 2 · 1.127 ms (−11.3%) | 183 · 1.760 ms (+38.4%) |
| home | 14 sheets · 1.079 ms | 2 · 1.038 ms (−3.9%) | 175 · 1.564 ms (+44.9%) |

The penalty persists when the sheets are already constructed and only matching is timed (+24.9% / +33.2% for `atomized`), so it is matching cost, not parse cost. Slope ≈ 1.5–2 µs per stylesheet per full-document restyle.

**Scale honestly:** at today's 7–15 sheets the recalc component is 0.04–0.18 ms — real, reproducible, correctly signed, and small. **The dominant win is network** (6 fewer render-blocking requests + 1,927 B gzip, both on the FCP/LCP critical path). The recalc measurement matters mainly because it establishes there is *no tradeoff* to weigh: grouping does not cost render time.

For scale: grouping saves more bytes (1,927 B) than removing *all* Light DOM scoping markers would (1,596 B), and unlike that option it costs nothing in correctness.

### Who would benefit?

Every app rendering more than a couple of components per route — i.e. essentially all non-trivial WebUI apps. The benefit grows with component count per route, so it helps most exactly where apps are largest.

### Desired outcome

**A `--css-bundle` flag that composes with `--css`, not a fourth `CssStrategy` variant.**

`CssStrategy` (`crates/webui-parser/src/lib.rs:165`) is a *delivery mechanism*: `Link` (external files), `Style` (inline `<style>`), `Module` (importmap data-URI + adopted stylesheets). Bundling is a *grouping* concern and is orthogonal to all three:

| | today | with `--css-bundle` |
|---|---|---|
| `--css link` | N `<link>` tags | 1 route `<link>` + shared chunk `<link>`s |
| `--css style` | N `<style>` blocks | 1 `<style>` block |
| `--css module` | N importmap entries | 1 entry per route + shared chunks |

This is not theoretical: **the measurement above was taken on `commerce`, which builds with `--css module`.** Grouping 15 inline sheets to 2 produced the −11.3%. Folding `bundle` into the delivery enum would make it impossible to express "module delivery, grouped" — which is exactly the configuration that was measured.

**Plus shared-chunk splitting.** Grouping strictly per-route duplicates any component used by more than one route. Routes sharing a component set should get a shared chunk, so `<mp-navbar>` is downloaded and parsed once across the whole app rather than once per route bundle.

### Concrete example

```bash
webui build ./src --css link --css-bundle --asset-file-name-template "[name]-[hash].[ext]"
```

```html
<!-- today -->
<link rel="stylesheet" href="/css/mp-navbar.css">
<link rel="stylesheet" href="/css/mp-product-card.css">
<link rel="stylesheet" href="/css/mp-price.css">
<!-- ...4 more... -->

<!-- with --css-bundle: shared chunk + route-local remainder -->
<link rel="stylesheet" href="/css/chunk-mp-navbar-a3f9c1.css">
<link rel="stylesheet" href="/css/route-product-7b2e04.css">
```

### Constraints

**Reuse the existing component-asset chunker — do not write a second one.** `crates/webui/src/component_assets/` already implements exactly this algorithm for ESM assets, and the doc comment on `render_component_assets` already states the rule this issue needs: *"Components used by one requested root stay inline, while components with an identical multi-root consumer set are emitted once in a shared chunk."*

Directly reusable:

- **`graph.rs::plan_component_assets`** — builds a `component × root` consumer bitset (`Vec<u64>`), classifies each component as unused / root-local / shared, sorts shared components by consumer row so identical sets are adjacent, then `group_shared_components` collapses each run into one `ChunkPlan`. This *is* the chunking algorithm; it needs no CSS-specific logic.
- **`traversal.rs`** — `GraphIndex` plus `TraversalScratch::collect`, an iterative (explicit-stack, no recursion) reachability walk with generation-marked scratch reused across roots, so N roots cost no extra allocation.
- **`entry_mask`** — components reachable from the entry are excluded from chunking as already-loaded prerequisites. CSS needs the identical concept for the app-shell sheet.
- **`AssetFileNameTemplate`** (`crates/webui-parser/src/asset_filename.rs`) — already supports `[name]`, `[hash]` (SHA-256 truncated to 8 hex), `[ext]`. **Content hashing does not need to be built.**
- **`validate_unique_asset_file_names`** and **`metafile::render_metafile`** (esbuild-compatible) — reusable as-is.
- **`ComponentStyleClosure`** in the protocol already models per-component style dependencies, and `retain_entry_protocol` already prunes it.

**The one genuine blocker.** `traversal.rs::enqueue_dependency` hard-errors on routes:

```rust
Some(Fragment::Route(_)) => return Err(routes_unsupported(owner)),
```

(diagnostic `COMPONENT_ASSETS_WITH_ROUTES`, help: *"routes cannot be combined with static component assets"*). Route-level CSS bundling is *entirely about routes*, so `plan_component_assets` cannot be called with routes as roots today.

This makes the shape of the refactor concrete — a three-way split rather than a straight reuse:

1. **Chunking core** — consumer bitset + `group_shared_components`, generic over opaque "unit" and "consumer" ids. Knows nothing about ESM, CSS, or routes. Pure and independently testable.
2. **Traversal policy** — two callers over shared `GraphIndex`/`TraversalScratch`: component-asset roots (rejects routes, preserving today's diagnostic) and CSS routes (walks them).
3. **Renderer** — ESM module emission vs CSS concatenation.

Only (1) is shared verbatim today; (2) is where the route restriction lives and must be parameterized rather than deleted, so the existing `COMPONENT_ASSETS_WITH_ROUTES` diagnostic keeps firing for ESM assets.

Other constraints:

- **Cache granularity is the real tradeoff and belongs in the docs.** A bundle invalidates on any component edit; 7 files invalidate one. Content hashing plus cold-cache first visits (what FCP/LCP grade) should outweigh it for most apps, but it's a genuine regression for apps shipping small frequent CSS edits to returning users. This is why `--css-bundle` should be opt-in rather than the default. Shared-chunk splitting also *mitigates* it: stable shared components stay in a stable chunk that survives route-local edits.
- **Cannot merge across a shadow boundary.** A component opting into Shadow via `<template shadowrootmode="open">` owns its own tree and needs its own sheet. The measured `bundled` arm above is 2 sheets, not 1, for exactly this reason.
- **Cascade order must be preserved exactly.** Today's order is deterministic via `document_style_resources` insertion order (`crates/webui-handler/src/lib.rs:1370`); reordering would be a silent visual regression. Worth an equivalence test asserting computed styles are unchanged before/after bundling — the same technique used to validate the measurement above.
- Orthogonal to Light DOM marker stamping; no interaction with `:where([data-wl-*])` qualifiers.
- **`emit_route_style_preloads` must be generalized to chunks, not removed.** CSS `<link>` tags are written into the streaming writer *at the point each component's style closure is emitted* (`crates/webui-handler/src/lib.rs:1398`) — they are not hoisted into `<head>`. A component rendering deep in the body therefore has its stylesheet discovered late, and `emit_route_style_preloads` exists to hoist that discovery into the head for every tag in the entry + route-chain closure not yet written. **Bundling changes how many files exist, not where in the document they are referenced, so the late-discovery problem survives it completely.** Two changes are needed:
  - Preload per **chunk href** rather than per component tag, deduping against emitted chunks instead of `document_style_resources` component tags.
  - A route under bundling still fetches its route-local bundle *plus* every shared chunk it consumes, so there remain multiple requests to overlap — and each hint now covers a whole chunk of CSS instead of one component's, so the payoff per hint goes **up**. Shared chunks are the highest-value preload targets in the app: consumed by several routes, most likely to be cache-warm, and referenced from whichever consuming component happens to render first.

### Alternatives or workarounds

- **A fourth `CssStrategy::Bundle`** — the original framing of this issue, and wrong: it conflates grouping with delivery and cannot express `module` + grouped, which is the configuration actually measured.
- **Whole-app bundle** — simpler, but ships unused CSS and regresses small routes. Shared-chunk splitting is the better form of the same idea.
- **Strict per-route bundles, no shared chunks** — duplicates common components across every route bundle, trading request count for payload and cache churn.
- **HTTP/2 multiplexing makes request count free** — reduces connection overhead, but the sheets stay render-blocking and it does nothing for the 38% compression loss.
- **Rely on an external bundler/CDN** — pushes a framework-level concern onto every consumer, and it can't see route→component mapping without duplicating the projection manifest.

mohamedmansour:
## Current behavior

WebUI uses Shadow DOM for unwrapped components by default. This PR makes explicit `--dom=light` a first-class mode for authored/global Light CSS:

- Unwrapped components under `--dom=light` render directly into the host and use ordinary CSS in their owning `Document` or `ShadowRoot` CSS tree.
- A sole bare top-level `<template>` is an explicit Light-mode wrapper even when the fallback is `--dom=shadow`; WebUI removes the wrapper before rendering and hydration.
- A sole `<template shadowrootmode="open">` explicitly selects native Shadow DOM in either build mode.
- Templates with attributes, `w-render`/`w-hydrate` policy wrappers, and nested templates do not select a mode; they retain their ordinary/policy meaning.
- Light CSS is not selector-rewritten, marker-scoped, wrapped in compiler-generated `@scope`, or namespaced for keyframes/layers. Normal cascade and inheritance apply across Light components in the same CSS tree.
- `:host`, `:host(...)`, `:host-context(...)`, and `::slotted(...)` in effective Light CSS fail at build time with `unsupported-light-css` and actionable help to use a normal selector or opt into Shadow.
- Ordered CSS closures remain the delivery contract across SSR, routes, streaming, partial navigation, hydration, and component assets. Link/Style/Module delivery, compatible bundling, active Document head hoisting, and ShadowRoot preloads are preserved.
- FAST 2/3 continues to require effective Shadow components; the WebUI plugin supports global Light CSS.

## Benchmark: representative 432-instance fixture

Eight browser cycles and eight release-server cycles were run on the current head with 48 component types, nested Light/Shadow variants, shared selectors, container queries, and bundled/unbundled Link delivery. Computed-style parity was identical across all 48 component types.

| Metric | Light | Shadow | Light delta |
|---|---:|---:|---:|
| Release render, unbundled | 0.09 ms | 0.12 ms | **-29.5%** |
| Release render, bundled | 0.08 ms | 0.12 ms | **-30.5%** |
| SSR output | 1.09 MiB | 1.15 MiB | **-5.1%** |
| Protocol, unbundled | 140.91 KiB | 144.06 KiB | **-2.2%** |
| Bundled DOMContentLoaded | 39.85 ms | 546.60 ms | **-92.7%** |
| Bundled stylesheet requests | 1 | 48 | **-97.9%** |
| Bundled style recalculation | 4,170.52 ms | 849.03 ms | +391.2% |
| Bundled host class workload | 317.05 ms | 139.90 ms | +126.6% |
| Bundled host attribute workload | 354.85 ms | 178.95 ms | +98.3% |

Global Light improves server rendering, output size, stylesheet consolidation, and document readiness. Native Shadow remains the stronger CSS-engine boundary for CSS-heavy or frequently restyled trees; this PR does not claim Light is universally faster.

## Validation

- `cargo xtask check`
- 601 parser tests
- 273 framework unit tests
- 222 framework E2E tests
- 28 router E2E tests
- 58 Node package tests
- 24 WASM tests

Closes #410
Closes #433


packages/webui-router/src/templates.ts
```diff
@@ -131,15 +109,54 @@ export function registerTemplatesAndStyles(
...
   }
+  const resources = { ...current.resources };
+  const closures = { ...current.closures };
+  for (const id of Object.keys(next.resources)) {
+    const existing = resources[id];
+    if (existing && JSON.stringify(existing) !== JSON.stringify(next.resources[id])) {
+      throw new Error(`[Router] Conflicting component style resource "${id}".`);
+    }
+    resources[id] = next.resources[id];
+  }
```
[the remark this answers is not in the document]
mohamedmansour: Fixed in c2fa6a36. The fallback merge now compares each resource discriminant and its relevant fields directly, so semantic equality no longer depends on property insertion order; a reordered Module-resource regression covers this path.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

packages/webui-framework/src/component-asset/loader.ts
```diff
@@ -106,20 +113,75 @@ function importAndPrepareChunk(
...
+    for (const component of prepared.asset.components) provided.add(component);
+    const styles = prepared.componentStyles;
+    for (const id of Object.keys(styles.resources)) {
+      const key = JSON.stringify(styles.resources[id]);
+      const current = resourceKeys.get(id);
+      if (current !== undefined && current !== key) {
+        throw new Error(`[WebUI] Conflicting component style resource "${id}".`);
+      }
+      resourceKeys.set(id, key);
+    }
```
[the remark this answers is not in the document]
mohamedmansour: No code change is needed in this path: every asset passes through requireComponentStyles and prepareComponentStyles before validatePreparedGraph, and prepareComponentStyles reconstructs each resource in canonical field order. Source property order therefore cannot affect these JSON.stringify keys; the separate router fallback path without that normalization was fixed in c2fa6a36.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

packages/webui-router/src/router.ts
```diff
@@ -57,6 +57,32 @@ const WEBUI_DATA_ID = 'webui-data';
...
 let webuiDataLoaded = false;
 
+function isRouteStyleMarker(node: Node): node is Element {
+  if (node.nodeType !== 1) return false;
+  const element = node as Element;
+  if (element.getAttribute('data-webui-resource') === null) return false;
+  const strategy = element.getAttribute('data-webui-strategy');
+  return (element.localName === 'link' && strategy === 'link') ||
+    (element.localName === 'style' && (strategy === 'style' || strategy === 'module'));
+}
```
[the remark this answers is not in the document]
mohamedmansour: Good catch on the router logic, but this case isn't reachable — the premise about where importmaps land is off.

`emit_component_style_closure` is the only path that writes into a route element, and it emits exactly two shapes: `<link ... data-webui-strategy="link">` and `<style ... data-webui-strategy="style|module">`. Under `CssStrategy::Module` the *routed closure* still emits a `<style data-webui-strategy="module">` fallback, not an importmap. The importmap comes from a separate call (`emit_css_module_importmap`) that writes inside the component's own Light DOM, so it is a descendant of the route host, never a direct child of the route element. `clearRouteContent()` and `mountedRouteComponent()` only walk direct children, so they never see it — and removing the outgoing host along with its nested importmap is the intended behavior.

Your reasoning about the consequences is right though: if an importmap ever did become a direct route child, `isRouteStyleMarker` would return `false` for it (`localName === 'script'`), it would be cleared on navigation, and it would be returned as the mounted component. That is an unguarded assumption, so I pinned the position rather than leaving it implicit. `test_css_module_emitted_for_route_components` now asserts the importmap offset falls after the `<dash-page data-wl>` host open tag, which fails if emission ever moves it out to route level.

Fixed in 5e361328.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

mcritzjam (dismissed):

crates/webui/src/style_bundle.rs
```diff
@@ -0,0 +1,762 @@
...
+    for tag in component_tags {
+        let css = css_by_tag
+            .get(tag.as_str())
+            .ok_or_else(|| missing_css(tag))?;
+        if !bundled.is_empty() && !bundled.ends_with('\n') {
+            bundled.push('\n');
+        }
+        bundled.push_str(css);
+    }
+    Ok(bundled)
```
KurtCattiSchmidt, line 459: CSS `@import` needs to be first, so you can't concat like this. But it shouldn't be used here anyways, as it's slow. Is `@import` explicitly blocked in WebUI2 CSS? It doesn't work with `adoptedStyleSheets` anyways, so it should probably be blocked.

crates/webui/src/style_bundle.rs
```diff
@@ -0,0 +1,762 @@
...
+///
+/// Excluding a closure is always safe — it keeps an empty `style_chunks` and
+/// every consumer falls back to per-component delivery, exactly as an unbundled
+/// build does. Including one that is never installed as a unit is not, so this
+/// stays conservative.
+///
+/// The entry comes first because chunks are shaped by whichever closure claims
+/// their members first and the entry is what every request pays for. Remaining
+/// roots are sorted so a build is reproducible.
+fn collect_roots<'a>(protocol: &'a WebUIProtocol, entry: &'a str) -> Vec<&'a str> {
```
KurtCattiSchmidt, line 355: Not sure if this is a real issue, please verify though:

"Excluded Light closures can receive unrelated bundled CSS

 crates/webui/src/style_bundle.rs  (lines 325–341)

Description:  collect_roots()  intentionally excludes ordinary Light component closures from the consumer matrix. However, components from those excluded closures are later resolved through the global component-to-chunk index.

For example:

index.html closure: [a-card, b-card]
lazy-panel closure: [a-card]       // ordinary Light component, not a planning root

The planner can merge  a-card  and  b-card  because only  index.html  is considered. When  lazy-panel  is independently requested, its  a-card  member resolves to that merged chunk, delivering both stylesheets.

Impact:  b-card  CSS is installed in a CSS tree that never requested it. Global selectors, layers, custom properties, and keyframes from that stylesheet can affect unrelated content. This contradicts the file’s stated invariant that no tree receives rules it would not receive unbundled.

Fix: Either include every independently deliverable closure in the consumer matrix, or map an excluded closure member to a chunk only when that chunk’s complete ordered membership is present in the closure. Otherwise, deliver the component’s individual fallback resource. Add a partial-overlap test like  [a-card]  against an entry chunk  [a-card, b-card] ."

crates/webui/src/style_bundle.rs
```diff
@@ -0,0 +1,762 @@
...
+}
+
+/// Break apart any chunk whose members are not contiguous and identically
+/// ordered in every closure that consumes it.
+///
+/// A chunk is contiguous in a closure exactly when walking that closure visits
+/// the chunk in one uninterrupted stretch. Member positions also have to advance
+/// from zero in the chunk's canonical order; a different relative order would
+/// silently change the cascade after concatenation.
+fn split_incompatible_chunks(
```
KurtCattiSchmidt, line 240: Compatibility checking reallocates a chunk-sized buffer for every root

Severity: 🟡 Performance
 crates/webui/src/style_bundle.rs  (lines 221–263)

Description:  split_incompatible_chunks()  creates this allocation inside the root loop:

```
for root in roots {
    let mut current = usize::MAX;
    let mut touched = Vec::with_capacity(chunks.len());
    // ...
}
```

The allocation is released and repeated for every root, even though only its length needs resetting.

Impact: A route-heavy application with thousands of roots and chunks performs substantial avoidable allocator traffic. In the worst shape, cumulative allocation volume approaches  O(roots × chunks)  even though the algorithm already has reusable per-chunk scratch arrays.

Fix: Allocate  touched  once before the loop and call  clear()  after resetting its entries:

```
let mut touched = Vec::with_capacity(chunks.len());

for root in roots {
    touched.clear();
    // existing walk and reset
}
```

crates/webui-handler/src/css_module.rs
```diff
@@ -30,17 +31,21 @@ pub fn build_importmap_tag(specifier: &str, css: &str, nonce: Option<&str>) -> S
     let data_uri = build_data_uri(css);
     let body = build_importmap_json(specifier, data_uri);
 
-    // `<script type="importmap"></script>` is 33 chars; `nonce=""` adds 8 +
-    // the value. A few extra bytes avoid a reallocation when the body is
-    // small.
-    let cap = 40 + body.len() + nonce.map_or(0, |n| n.len() + 9);
+    let escaped_specifier = crate::html_encode::encode_safe(specifier);
+    // `<script type="importmap" data-webui-resource=""></script>` is 56 chars;
```
KurtCattiSchmidt, line 35: The new import-map marker is not skipped during hydration

Severity: 🔴 Bug
Files:  crates/webui-handler/src/css_module.rs  (lines 34–48),  packages/webui-framework/src/element/markers.ts  (lines 218–230, 305–307)

Description: The branch adds  data-webui-resource  to the generated  <script type="importmap"> , explicitly so Light DOM hydration excludes it from template node ordinals. However, hydration skips elements only when  isComponentStyleMarker()  returns true. That function recognizes only:

```
<link data-webui-resource data-webui-strategy="link">
<style data-webui-resource data-webui-strategy="style|module">
```

It does not recognize import-map scripts, and the generated script has no  data-webui-strategy  attribute anyway.

Impact: When an import-map script is emitted before a component in Light DOM,  buildSSRIndex()  can pair the compiler-generated  <script>  with the authored component’s template index. Attribute bindings and events intended for the component may then be wired to the script, shifting all subsequent element indexes.

The Light-first behavior introduced by this branch makes the problem substantially more exposed.

Fix: Add a distinct compiler-resource predicate that recognizes:

```
<script type="importmap" data-webui-resource="...">
```

Use it for SSR ordinal exclusion, but do not necessarily treat it as a retained style element during reconnect. Add a hydration test with a module import map followed by elements with bindings/events.

crates/webui/src/style_bundle.rs
mohamedmansour: Fixed. A non-root Light closure now resolves a bundled chunk only when the chunk's complete ordered membership is present; partial independent closures fall back to their component resources. Added a regression test for the `[a-card]` versus `[a-card, b-card]` case.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

crates/webui-handler/src/css_module.rs
mohamedmansour: Fixed. Added a distinct compiler style-resource marker predicate for `script[type="importmap"][data-webui-resource]`; SSR ordinal lookup skips it, while retained reconnect styles continue using the Link/Style predicate. Added a `buildSSRIndex` regression test.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

crates/webui/src/style_bundle.rs
mohamedmansour: Fixed by allocating the `touched` scratch vector once outside the root loop and clearing it between roots. The full Rust gate passes with the change.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

crates/webui/src/style_bundle.rs
mohamedmansour: Good catch. Bundling now rejects top-level `@import` rules with an actionable error because concatenation cannot preserve import ordering; unbundled CSS remains supported. Added coverage for imports, comments, strings, and declaration values.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

janechu (dismissed):

janechu (approved):

PR #450, opened 2026-08-18
fix: scale FAST route state with scalar attributes

janechu:
## Summary

- expose FAST v2 and v3 route state through escaped scalar kebab-case attributes
- share route-state attribute emission across both FAST handler plugins
- document scalar FAST state initialization and built-in plugin identifiers

## Why

FAST route state uses lightweight scalar attributes for `@attr` inputs. Complex component values initialize from rendered DOM or application-specific loading paths, keeping SSR response size and allocation cost proportional to the scalar attribute surface.

## Dependency

- #451 restores the repository quality gate and is intentionally isolated from this FAST change

## Validation

- `cargo xtask check`
- `cd docs && pnpm build`
- framework code review: no findings


crates/webui-handler/src/plugin/mod.rs
```diff
@@ -6,6 +6,8 @@
 //! Handler plugins write framework-specific hydration markers while shared
 //! completion work, such as component template emission, stays in handler core.
 
+mod fast_route_state;
```
mohamedmansour: shouldn't this go in fast / fast_v2 / fast_v3 since it is dependent on that

docs/ai/SKILL.md
```diff
@@ -911,7 +911,10 @@ identifier is falsy, so `path` is false and `!path` is true. No error.
 
 **Route-scoped state.** Each route handler should return only the keys that
 route's template binds to. Sending full app state on every route wastes
-bandwidth and render time.
+bandwidth and render time. FAST route components expose scalar values as
+kebab-case HTML attributes, so keep `@attr` inputs flat. Strings, numbers, and
+booleans are emitted. Complex values should be initialized from the rendered
+DOM or another documented mechanism appropriate to the component.
```
mohamedmansour: we don't care about how the internals work in the docs, so this can be removed

docs/guide/concepts/plugins/index.md
```diff
@@ -33,6 +33,21 @@ webui serve ./my-app --state ./data/state.json --plugin=<name>
...
+| `fast` | Deprecated alias for `fast-v2` |
+| `fast-v2` | Deprecated FAST 2 compatibility name |
+| `fast-v3` | FAST 3 hydration plugin |
+| `webui` | WebUI framework hydration plugin |
+
+FAST plugins expose scalar route state as kebab-case HTML attributes on
+matched route components. Strings, numbers, and booleans are emitted.
+Complex values should be initialized from the rendered DOM or another
+documented mechanism appropriate to the component. Keep route payloads flat
+when the component reads them through `@attr`.
```
mohamedmansour: we can remove this text, and just keep the above table only

deny.toml
```diff
@@ -2,8 +2,6 @@
 ignore = [
```
mohamedmansour, line 2: revert this file

mohamedmansour (approved):

PR #454, opened 2026-08-19
fix: prevent dynamic component stylesheet flashes

mohamedmansour:
Dynamic Link-mode components could become visible before their stylesheet was ready during lazy creation or soft navigation. This change keeps the browser's native stylesheet link authoritative, prevents unstyled paint on the first client-created mount, and reuses an authorized constructable sheet for later mounts.

## High-level design

The solution covers the three ways WebUI content reaches the browser:

- **Full SSR page loads:** continue using parser-discovered native stylesheets and declarative shadow roots. Hydration preserves the server-rendered DOM.
- **Router navigation:** JSON, streaming, and `Router.ensureLoaded()` template registration now expose a readiness handshake. The framework prepares Link styles before the router commits the new component, while stale navigations can still abort promptly.
- **Static component assets:** the compiler records final generated CSS hrefs for each asset root and its shared closure. The protocol carries that metadata to the handler, which publishes it for the existing `defineComponentAssets().preload(tag)` lifecycle. CSS, the root asset, the component module, and optional data can therefore start in parallel without application code knowing generated hashes.

For a first client-created Shadow DOM instance, the framework:

1. Uses `<link rel="preload" as="style">` only to warm the matching style request.
2. Keeps the original `<link rel="stylesheet">` authoritative for CSP, MIME, integrity, CORS, redirects, and service workers.
3. Guards the host until the native link is active, preventing unstyled paint.
4. Releases the guard immediately, then promotes browser-authorized CSSOM into a shared constructable sheet when safe.
5. Falls back to the original native links whenever promotion is unsupported or ambiguous.

Light-DOM component assets receive real, deduplicated document stylesheets because global CSS must be applied rather than merely preloaded. Failed asset or authored-module loads are evicted so a later preload or create call can retry.

## Developer experience

There is no public raw-href API and no generated CSS filename in application code:

```ts
const assets = defineComponentAssets({
  'lazy-panel': {
    asset: './lazy-panel.webui.js',
    module: () => import('./lazy-panel.js'),
  },
});

assets.preload('lazy-panel');
panel.replaceChildren(await assets.create('lazy-panel'));
```

Bundlers may instead own the root asset import:

```ts
asset: () => import('./lazy-panel.webui.js')
```

Router-owned components continue using router template loading; component assets remain for non-routed deferred UI emitted through `--emit-component-assets`.

## Measured impact

Independent Edge NTP retest of `<ntp-waffle-menu>`, with a fresh browser context per run, 4x CPU throttle, and 150 ms RTT:

| Scenario | 0.0.22 baseline | Final | Change |
|---|---:|---:|---:|
| Cold click to styled paint | 469.8 ms | 241.4 ms | -49% |
| Warm reopen | 222.3 ms | 25.1 ms | -89% |
| 50 ms intent dwell | - | 167.9 ms | CSS and JS start in parallel |

Every tested path made exactly one CSS request. Cold and warm runs produced 0/10 unstyled first paints, and constructable-sheet adoption succeeded in 10/10 cold runs. Instrumentation placed the framework CSS request start within about 1.6 ms of a directly injected preload; compiler-manifest parsing was not measurable.

Against `origin/main` in the 500-component responsiveness harness:

- Unstyled visible frames: 8 -> 0
- Cold styled paint: 140.38 -> 124.74 ms
- Warm mount: 6.25 -> 3.815 ms
- Main task time: 20.905 -> 16.975 ms
- Heap: +9,976 bytes total, about 20 bytes per component
- Backing storage: -169,200 bytes

JavaScript footprint is +4,514 gzip bytes in the main framework entry versus `origin/main`; the optional component-asset entry adds 271 gzip bytes versus the prior PR head.

## Important behavior

- Constructable-sheet promotion is a progressive enhancement. Native links remain the fallback.
- A stylesheet error leaves a client-created component guarded instead of exposing unstyled content.
- Automatic early component-asset style preloading requires handler or `Protocol` rendered HTML containing the compiler-owned manifest. Using build artifacts without rendering that protocol remains paint-safe, but CSS starts after the root asset reveals its template metadata.
- Intent that never mounts may produce the browser's standard unused-preload warning before bounded cleanup removes the speculative node.

## Validation

- Full `@microsoft/webui-framework` unit and Playwright suites
- Router unit and E2E coverage for JSON, streaming, stale navigation, and no-framework behavior
- Component asset compiler, protocol, handler, Shadow/Light, and retry regressions
- Developer documentation build
- `cargo xtask check`


packages/webui-router/src/streaming.ts
```diff
@@ -91,32 +109,107 @@ export async function readStreamingPartial(
...
+  const data = chunk1 as StreamingPartialResponse;
+  let consumed = false;
+  const control: DeferredStreamControl = {
+    async cancel() {
+      if (consumed) return;
+      consumed = true;
+      deferredStreams.delete(data);
+      await cancelReader(reader);
+      ctx.setDeferredReader(null);
+    },
```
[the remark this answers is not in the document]
mohamedmansour: Fixed in f2c8ed39. Both cancellation and consumption now remove `_deferredStream` immediately, with regression assertions covering both paths so cached responses no longer trigger the lazy import after ownership ends.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

packages/webui-framework/src/element/link-styles.ts
```diff
@@ -0,0 +1,1060 @@
...
+    const result =
+      state.descriptors.length === links.length
+        ? state.results?.[index]
+        : undefined;
+    const href = result?.href || links[index].href;
+    let reason = '';
+    if (result?.reason) {
+      reason = ` Constructable preparation failed: ${result.reason}.`;
+    }
+    console.error(
```
Qusic, line 199: Traced this to be sure: the guard is only released in `finishLoadedMount()`
   (L310) and `cancel()` (L327). `fail()` releases nothing, and
   `installTemplateLinkStyles` runs once per mount (`$hydrated` gate,
   template-element.ts:586), so nothing retries. The only exit is removing the
   element from the DOM so `$destroy()` fires, then recreating it.

   It lands in two shapes, and the second is the worse one:

   - Guard effective: content is appended, host stays `visibility:hidden`. The
     component is fully wired and interactive, just invisible — and since
     `visibility:hidden` keeps the box, it's a correctly-sized blank hole rather
     than a collapsed layout.
   - CSP blocks the guard style (`deferContent`): content never leaves staging,
     and `afterAppend` only runs from `settle()`, so `$finishHydration()` never
     fires and `$ready` stays `false` permanently.

   On `main` today the same stylesheet failure gives an unstyled but visible and
   usable component, so this is a new state rather than a changed one.

   What I'm unsure about is the host application's position: the only signal is
   `console.error`, and there's no way to act on it short of destroying and
   recreating the element itself. On Edge NTP this covers the client-created
   surfaces — sign-in flyout, settings, market/language dialog — so a user clicks
   "Sign in" and nothing appears.

   Is permanent invisibility the intended end state here, and how would you want
   an app to handle it?

mohamedmansour: Good catch - permanent invisibility was not an acceptable failure mode. A definitive native link error now logs the failure, leaves the native links authoritative, releases the temporary guard, and completes any CSP-deferred append/hydration so the component remains visible and usable (though potentially unstyled); browser tests cover both guard shapes.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

Qusic (approved):

PR #481, opened 2026-08-25
perf: keep routed template closure counts exact

mohamedmansour:
## Summary

- keep `Symbol.for('microsoft.webui.templateFnCount')` exact when router partials add or replace template closure arrays
- reset the symbol count when `Router.gc()` clears the registry's string keys
- cover late writes, replacement, repeated soft navigation, GC reuse, and deletion exactly at zero

## Why D1 is required

A8 lazily initializes the closure-registry count in the framework. A router partial written after that initialization previously added a string key without incrementing the symbol count (the regression test observes `1` where `2` is required). `Router.gc()` also removed the string keys while leaving the initialized count stale. After a 50-entry registry, one normalization, GC, and 25 soft navigations, A8 retained an empty registry with count `24`; D1 deletes the registry after each consumed post-GC write.

The router computes the next count synchronously before executing its nonce-gated closure script, then adds only the final numeric symbol assignment to the generated script. This avoids the larger per-tag counter code from the initial implementation.

## Measurements

Chromium 149, compared with A8 head `1d12f1b2`. Timing uses 24 measured samples after 4 warmups and 100 partial registrations per sample.

| Metric | A8 | D1 | Delta |
|---|---:|---:|---:|
| Generated script, 1 tag | 147 B | 198 B | +51 B |
| Generated script, 5 tags | 331 B | 382 B | +51 B |
| Generated script, 10 tags | 561 B | 613 B | +52 B |
| Minified browser benchmark bundle | 48,799 B | 49,145 B | +346 B |
| Startup median / p95 | 1.1 / 1.3 ms | 1.0 / 1.3 ms | no regression |
| Registration median, no pending closures | 0.013 ms | 0.014 ms | +0.001 ms |
| Registration median, 50 pending closures | 0.011 ms | 0.012 ms | +0.001 ms |
| Registration p95, 50 pending closures | 0.016 ms | 0.014 ms | -0.002 ms |
| Warmed post-GC retained-heap delta | 43,700 B | 43,940 B | +240 B |

The forced-GC retention scenario seeds 50 closure entries, normalizes one, calls `Router.gc()`, then performs 25 realistic partial registrations. Retained templates stay equal at 25 and retained closure arrays stay at 0; the feature-specific registry-object count improves from 1 to 0 and the stale symbol value changes from 24 to absent. The +240 B total-heap difference is allocator-level overhead for the exact-count path and is bounded, while the stale registry behavior is eliminated.

## Validation

- 94 router unit tests passed
- 31 router browser E2E tests passed; 4 existing Navigation API cases skipped
- framework code review found no significant issues
- `cargo xtask check` passed on current `main`

### DIFF
--- a/packages/webui-router/src/router.ts
+++ b/packages/webui-router/src/router.ts
@@ -353,6 +353,9 @@ export class WebUIRouter {
       for (const tag of Object.keys(functionRegistry)) {
         delete functionRegistry[tag];
       }
+      const counts = functionRegistry as Record<string, unknown>
+        & Record<symbol, number | undefined>;
+      counts[Symbol.for('microsoft.webui.templateFnCount')] = 0;
     }
     if (window.__webui) window.__webui.inventory = '';
   }
--- a/packages/webui-router/src/template-function-count.test.ts
+++ b/packages/webui-router/src/template-function-count.test.ts
@@ -1,0 +1,175 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import './browser-shim.js';
+
+import { strict as assert } from 'node:assert';
+import { test } from 'node:test';
+
+import { WebUIRouter } from './router.js';
+import { registerTemplatesAndStyles } from './templates.js';
+
+const TEMPLATE_FN_COUNT = Symbol.for('microsoft.webui.templateFnCount');
+
+type TemplateFunctionRegistry = Record<string, unknown>
+  & Record<symbol, number | undefined>;
+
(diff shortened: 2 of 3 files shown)


mohamedmansour:
Closing as D1 NO-GO after reviewing the measurements against the strict performance bar. The change removes one empty registry object, but closure arrays were already released; it adds 51-52 generated bytes, 346 bundle bytes, 240 retained heap bytes, and about 1 microsecond median registration latency. The counter-correctness improvement alone does not justify a perf PR.

PR #494, opened 2026-08-26
fix: preserve FAST property bindings during hydration

janechu:
## Summary

- add end-to-end coverage for structured `:prop` SSR scope, object identity, and synchronous `setState` propagation
- accept FAST `<f-repeat positioning="true">` while retaining the option for the FAST client template and omitting it from the WebUI SSR `<for>` view
- document that WebUI complex properties provide SSR component scope only, while FAST components own client state and hydration readiness
- update the FAST todo example to initialize state through its component lifecycle and verify that SSR survives hydration without a WebUI state side channel

## Testing

- `cargo xtask check`
- `pnpm --dir examples/app/todo-fast test` - 16 passed
- FAST `main` nested property-binding regression using the current `Protocol.render()` API and direct `.template.html` plugin path - 1 passed
- `pnpm --dir docs build`

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1400,7 +1400,9 @@ handler.render(&protocol, &state, &options, &mut writer)?;
   - **Boolean (with `condition_tree`):** Evaluate condition; if truthy, render attribute name only. If false, omit entirely.
   - **Simple dynamic (with `value`):** Resolve signal from state, render as `name="resolved_value"`.
   - **Template (with `template`):** Render `name="`, process referenced sub-stream, render closing `"`.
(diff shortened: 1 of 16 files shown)


mohamedmansour (commented):
The tests in packages/webui-framework/tests/ don't make sense for this PR, you can remove them

janechu:
Addressed the review feedback in ee76b9c1: removed the unrelated additions under packages/webui-framework/tests/. The PR now keeps only the FAST-specific regression coverage.

mohamedmansour (approved):

janechu:
Not relevant to current needs.

PR #523, opened 2026-09-11
feat: add Press content mode and plugin-aware discovery

mohamedmansour:
Component galleries need a shell-free documentation view without losing server-rendered examples or hydration. This adds that presentation to native Press and makes component discovery follow the selected plugin's authoring conventions instead of assuming every npm package uses the same layout.

## Approach

- Add `--show=all|content` to Press build and serve, defaulting to `all`. Content mode omits shell markup, scripts, and layout constraints before compilation while retaining metadata, authored examples, page modules, SSR, and hydration. The CLI override survives config reloads, including home/custom/404 pages.
- Separate content and shell styles. Full-mode manual themes override OS preferences for native tokens and controls; content mode follows live OS changes without applying the saved shell preference. Forced-colors overrides retain precedence.
- Keep default discovery filename-based. Isolate FAST's CEM-driven inventory and standard `*.template.html` resolution, with ordinary HTML fallback for undeclared components. Support symlinked packages, ancestor package/scope lookup, and npm collection spellings; report broken declared components instead of silently omitting them.
- Make package metadata an explicit discovery opt-in. Default/WebUI/none skips package JSON loading, parsing, hashing, and FAST export/ownership analysis. FAST still supports ordinary named HTML/CSS/script components directly in the app folder.
- Preserve parser error chains, resolve config-relative aliases consistently, serialize concurrent cold asset extraction, and exclude Finder metadata. Move FAST-specific documentation beside its implementation.

## Validation

- `cargo xtask check`
- Discovery regressions for filename naming, CEM inventory, fallback, caching, symlinks, scoped lookup, and error propagation.
- FAST 2 and FAST 3 build/render coverage.
- Press browser coverage for content/all output, no-JS SSR, same-node hydration, config reloads, native themes, OS changes, forced colors, and existing shell behavior.
- Metadata-isolation regressions prove unused metadata is not parsed and does not invalidate filename-discovery caches.

An optimized synthetic probe (7 rounds of 200 warm discoveries, one unchanged component) measured a metadata-heavy 377 KB package at roughly 1.2-1.5 ms before and 0.2-0.3 ms after the metadata opt-in change. These are shared-machine measurements, not a general throughput guarantee.


mohamedmansour:
Superseded by two focused, stacked PRs: #524 for plugin-aware discovery and #525 for WebUI Press content mode. The combined tree preserves this PR's changes plus the requested FAST converted-template correction and module-folder layout.

PR #536, opened 2026-09-14
feat: use Trusted Types automatically in the framework

mohamedmansour:
## Summary

WebUI now uses Trusted Types automatically for native compiled templates and generated CSS import maps. The private, fixed-name `webui` policy is created on first use. There is no configuration API, setup entry point, special import order, or global policy state.

Applications choose enforcement through CSP:

```text
require-trusted-types-for 'script'; trusted-types webui
```

Policy creation failures are explicit; browsers without Trusted Types retain normal string rendering. Raw state HTML and arbitrary script/source strings are never promoted. Rejected raw-HTML updates preserve existing DOM and allow remaining queued updates to proceed.

Includes updated developer documentation and regression coverage. Router/CLI integration remains separate; no dependency or release changes.

## Validation

- Framework TypeScript compilation and **360 unit tests passed**.
- Browser fixture typecheck and **16 Edge browser tests passed**: automatic hydration/client mounts, no global trust capability, enforced versus unenforced raw HTML, policy-name denial with and without enforcement, and existing raw-HTML regressions.
- `cargo xtask check`: **all stages passed in 177.3s**, including docs.
- Used installed Edge with a temporary local test-server configuration; committed tests retain standard CI configuration.

## Size

Compared with the previous explicit-configuration implementation (`13b0a17e`), the standalone minified policy module shrank from **2,133 to 1,242 bytes** (42%). A production framework bundle with Trusted Types enabled shrank from **78,344 to 77,461 bytes** (gzip: 24,778 to 24,484).

Automatic support is now included for every app: relative to the previous bundle that omitted opt-in code, the cost is **555 minified bytes / 208 gzip bytes**. The policy and weak maps remain lazy; unsupported browsers allocate neither.

Scope: **18 files, +742/-33**.


packages/webui-framework/src/template-element.ts
```diff
@@ -1622,43 +1622,51 @@ export class TemplateElement extends HTMLElement {
...
-          wc.attrs,
-          wc.conds,
-          wc.repeats,
-          requireKnownState,
-        );
+    } finally {
+      // A rejected sink still propagates its error. Release scheduling and keep
+      // independently queued re-entrant writes, without retrying the failed batch.
+      this.$pendingFlush = false;
+      if (this.$dirtyPaths?.size) {
```
Qusic, line 1673: Could we preserve the unprocessed entries from the current `dirty` snapshot when a binding throws? `$dirtyPaths` has already been cleared, so if `raw` and `count` are queued before the same flush and the `raw` Trusted Types sink rejects, iteration exits and `count` is neither applied nor requeued. This can leave state and DOM permanently out of sync; the `finally` block only sees paths added re-entrantly after the snapshot. A regression test using `$update('raw'); $update('count')` reproduces this. Please process or requeue the remaining paths while excluding the failed path before propagating the error.

packages/webui-framework/src/trusted-types-policy.ts
```diff
@@ -0,0 +1,136 @@
...
+      body += '})();';
+      const script = document.createElement('script');
+      if (nonce) script.nonce = nonce;
+      script.textContent = policy.createScript(body, capability) as string & TrustedValue;
+      document.head.appendChild(script);
+      script.remove();
+    },
+  };
+  // One boundary per document also serves split/duplicated module entrypoints.
+  Object.defineProperty(window, '__webuiTrustedTemplates', { value: Object.freeze(boundary) });
```
Qusic: Does exposing the complete boundary on `window` defeat the private-capability guarantee? These methods close over and supply `capability` for any caller, so application code can register an arbitrary `{ h }` object and pass it to `setTemplateContent()`. It can also pass arbitrary source and a nonce to `installTemplateFunctions()`. Under enforced Trusted Types in Edge, I verified that these calls can insert arbitrary TrustedHTML and execute arbitrary TrustedScript. `Object.freeze()` prevents mutation of the bridge but does not restrict invocation. This appears to contradict the stated absence of a generic HTML/script conversion API. Could the global state be limited to idempotency metadata while keeping trust-producing operations private, or otherwise require an unforgeable internal caller?

packages/webui-framework/src/template-element.ts
mohamedmansour: Fixed in cc724b01: a rejected path now requeues the unprocessed tail of the dirty snapshot alongside re-entrant writes, without retrying the failed or already processed paths. Added unit regressions for queued `raw`/`count`, multiple failures, and re-entrant writes, plus an enforced-Trusted-Types browser regression confirming the previous raw DOM survives while `count` updates.

packages/webui-framework/src/trusted-types-policy.ts
mohamedmansour: Fixed in cc724b01: removed the global callable bridge and compiler-source execution helper; policy operations now stay in module-private state, and the only global metadata is an immutable policy-name string. Independently bundled framework copies now fail with guidance to share one runtime instance; unit and enforced-Trusted-Types Edge regressions verify no global trust methods are exposed, ordinary HTML/script strings remain blocked, and compiled blocks still mount.

Qusic (dismissed):

Qusic (dismissed):

Qusic (approved):

PR #540, opened 2026-09-17
fix: forward HTML component inputs during SSR

mohamedmansour:
## Summary

A bare component input such as `<mai-drawer open aria-label="Canvas information">` remained on the host but did not supply the flag to the component's SSR state. ARIA inputs were also excluded from component scope.

This is now an HTML parser/renderer fix:
- Forward valueless component inputs as presence flags using existing boolean fragments.
- Preserve empty literal attributes and forward ARIA inputs through existing name mapping.
- Keep `?open="{{open}}"` behavior unchanged for true/false state.

The final diff is **five files**. The earlier JavaScript/decorator analysis and projection-manifest changes are removed. There are no protobuf, browser-runtime, package, or public-doc changes.

## Validation

- Native parser, handler and expression suites pass; Node package: 72 tests passed.
- The actual drawer renders open, visible and named with JavaScript disabled and **no projection manifest**.
- Unchanged actual drawer suite: 18 journeys run twice, **36/36 passed**, including literal-open/name/content and false-state paths. Only the native addon was replaced; installed compiler/runtime and component code were unchanged.
- Qualified addon: commit `7753b35874dc1d7e93861fa5b4cab76f3d858409`, SHA-256 `B2B86E644C7493957C994D3ACF34B504D15AD70EC7DB4A5022067FEAF3602806`.
- `cargo xtask check` passed in 154.1s using process-local TEMP/TMP on the worktree drive. The pre-existing Press default-temp cross-volume `PROJ-C013` docs-build failure remains separate and unmodified.


crates/webui-parser/src/lib.rs
atulkatti, line 4516: This may have some impact on protocol file size.

atulkatti (approved):

crates/webui-parser/src/lib.rs
mohamedmansour: Measured serialized protobuf bytes using identical parser fixtures at base `a357c26c` and head `7753b358`: one valued ARIA input (static, dynamic, or mixed) is unchanged, and three valued ARIA inputs shrink by 4 bytes because these were already attribute fragments and no longer encode `attr_skip`. There is a cost for previously raw bare inputs: the literal-open drawer fixture grows from 265 to 280 bytes (+15 per instance; +1,500 for 100 instances), and a bare ARIA input grows by 29 bytes in this fixture; those existing fragments now carry the component state needed for correct SSR, with no new protocol fields or schema.

PR #549, opened 2026-09-18
feat: add opt-in bounded dev-server shutdown

mddinbox:
## Summary

Opt-in follow-up to #546. That change makes normal dev-server shutdown wait for the active rebuild and discards queued work. This preserves that behavior by default and adds a separate process boundary for users who need to stop a rebuild that does not return.

Both native commands accept a positive integer `--shutdown-timeout <SECONDS>`:

```sh
webui serve ./src --watch --shutdown-timeout 10
webui-press serve --shutdown-timeout 10
```

It also applies to `webui serve` without `--watch` and to initial startup/build. No timeout is selected by default; the ordinary in-process path adds no supervisor or control thread.

## Design

This lives in the shared dev-server crate rather than importing `xtask/src/process.rs`: that binary-local helper has different best-effort containment/direct-child Unix termination semantics and no controlled startup gate. Extracting or changing it would widen this follow-up.

- One long-lived, same-executable server child retains the existing rebuild worker and warm cache. Original arguments and working directory are preserved.
- Startup is gated before extraction/build/output work. The parent releases it only after checked Windows Job assignment or Unix process-group creation. Containment failure does not fall back to uncontained execution.
- The first stop request stops HTTP, drops the watcher, and joins the existing rebuild worker. A second request or expired grace period terminates the owned process scope, not an arbitrary Rust thread. The monitor does not synchronously print progress before termination.
- Forced outcomes are non-success. The supervisor confirms scope termination before returning an ordinary outcome; confirmation errors explicitly do not mean it is safe to delete outputs. Normal child errors retain their exit code and are not printed twice, including `webui --format json`.
- Unix retains the unreaped group leader through the last group signal, avoiding a stale PID/PGID kill. A private successful-teardown status is emitted only after the controlled CLI returns from joining, and mapped to zero only after scope confirmation. It is not an arbitrary-program supervision API.

In opt-in mode all supported stop signals use Actix's non-draining Ctrl-C HTTP policy. In particular, opt-in SIGTERM does not use Actix's default graceful HTTP drain. Flag-absent behavior stays on the existing Actix signal path.

## Validation

These are local results for the integrated production implementation, not the earlier standalone prototype. The PR remains draft pending platform CI review.

- Windows x64: final pre-commit `cargo xtask check` passed (130.7 seconds), including license headers, formatting, workspace Clippy, protocol drift check, dependency audit, workspace tests/build, three WASM builds, example builds, benchmark smoke, and documentation build.
- Actual Windows binary integration tests: 4 for `webui`, 3 for `webui-press` passed. These cover gated startup, real HTTP serving/rebuild, normal joining, watch/no-watch, lost control, Press display-mode override, invalid policy, and preserving a single JSON error with exit code 66.
- Linux Docker with `--init`: shared dev-server tests passed (77 passed, 1 intentionally ignored subprocess-fixture entry); all 9 actual CLI binary tests passed. These include default SIGINT/SIGTERM/SIGQUIT and opt-in SIGINT/SIGTERM/SIGQUIT/SIGHUP for both commands.
- Shared-core process tests cover hung rebuilds, deadline/second-request escalation, descendant/grandchild writers, independent trees staying alive, unexpected root exit, startup gating, parent-control loss, and real Actix stop followed by active-work joining. Unix SIGQUIT forwarding/restoration is also covered.
- Linux release binaries: 12 real CLI runs (both commands, default/opt-in, three repetitions) served HTTP, rebuilt after an edit, and exited successfully on SIGINT. Process inspection observed one process without the flag, two with it, and unchanged server PIDs through rebuilds.
- Three-sample, tiny-fixture release measurements are descriptive only: summed process RSS was about 5 MiB higher in opt-in mode. Summed RSS can double-count shared pages and is not incremental physical/PSS memory. Rebuild wall times included watcher debounce; these are not throughput or latency guarantees.
- Production targets and the new CLI integration-test targets passed Clippy with `-D warnings`. An optional broader `--all-targets` run found existing disallowed `unwrap`/`expect` uses in unrelated test modules; those were not changed or suppressed. The repository's configured full gate passed. The shared shutdown crate passed strict all-target Clippy on Windows and Linux.

## Limits

- The grace budget is followed by up to two seconds of termination confirmation; it is not a hard real-time command-exit guarantee.
- OS-uninterruptible work, escaped/daemonized descendants, and force-killing the external Unix supervisor are outside the contract. Unix confirmation can remain unconfirmed when a host does not reap orphaned descendants.
- Forced termination can leave incomplete output. Rebuild before using it, and do not assume cleanup is safe after unconfirmed termination.
- Supervised mode reserves stdin and does not forward interactive input to build tools.
- macOS has not been run locally. Targeted native dev-server tests are added to the existing macOS and Windows PR build jobs; Linux already runs workspace tests. CI results must be reviewed before removing draft status.
- Windows Job containment, process/control-pipe behavior, and real child HTTP shutdown were tested natively. Interactive Windows console Ctrl-C/close delivery was not exercised end-to-end in this environment.


mohamedmansour:
### P2: Supervised startup can hang in Unix terminals with `TOSTOP`

In [`shutdown/unix.rs:20–22`](https://github.com/microsoft/webui/blob/b1c844dd4d97d9cb45aade7c669168b2a337822e/crates/webui-dev-server/src/shutdown/unix.rs#L20-L22), `command.process_group(0)` puts the worker in a separate process group, while [`Tree::spawn_attached`](https://github.com/microsoft/webui/blob/b1c844dd4d97d9cb45aade7c669168b2a337822e/crates/webui-dev-server/src/shutdown/tree.rs#L31-L36) leaves stdout/stderr inherited. The supervisor stays in the terminal's foreground group, so the worker is a background job from the terminal's perspective.

With `stty tostop`, the worker's first terminal write sends `SIGTTOU` to its process group and suspends it. There is no child-side handling or supervisor recovery for that stop. Both serve commands print startup output, so enabling `--shutdown-timeout` can prevent startup in this terminal configuration. The grace deadline does not start until a shutdown request arrives, so it does not automatically resolve this startup hang.

I reproduced this on Linux at `b1c844dd` using a small PTY probe that calls this PR's actual `shutdown::prepare` and `Control::serve` implementations with an Actix server:

| Mode | HTTP ready | Worker state |
| --- | --- | --- |
| Direct | Yes | `S (sleeping)` |
| Supervised | No | `T (stopped)` |

Please keep the separate process group for containment, but establish an explicit child-side `SIGTTOU` policy before the first application output, without changing the user's terminal settings. Add a PTY regression with `TOSTOP` enabled covering direct and supervised startup. This should be a focused platform fix, not a broader supervisor rewrite.

mohamedmansour (approved):

PR #552, opened 2026-09-20
fix: keep Press builds on one filesystem volume

mohamedmansour:
## Why

WebUI Press fails to build on Windows whenever the project checkout and `TEMP`/`TMP` live on different drive letters. Reported from `edge-microsoft/webui-components`: a gallery on `E:` with the default `C:\...\Temp` fails with

```
error: PROJ-C013: Adapter module graph is incomplete or inconsistent
```

The same checkout builds fine once `TEMP` is pointed at the project's drive, and a plain `git archive` extraction reproduces it with no git or worktree involvement.

## Root cause

Press extracts its embedded template and built-in components into `std::env::temp_dir()`. That extracted tree is not inert data: it contains real TypeScript sources (`template/index.ts`, `template/docs-search/docs-search.ts`, `components/*/*.ts`) that Press feeds to esbuild as the root bundle script and component sources.

The projection compiler stores every manifest key as `path.relative(rootDir, target)`, so one bundler invocation must resolve to a single build root. With inputs on `C:` and the manifest and outputs on `E:`, the adapter's common-ancestor scan walks to the filesystem root and finds nothing shared, then reports the generic `PROJ-C013`. This is not path normalization or a cross-volume `rename`; it is a genuine contract violation with a misleading error.

## The fix

**Keep the whole build on one volume.** New `crates/webui-press/src/scratch.rs` resolves a scratch base per project: the system temp directory when it shares the project's volume (so unrelated projects still share one cache), otherwise a self-ignoring `<config-dir>/.webui-press-cache`. Embedded-asset extraction plus the per-page and 404 scratch directories all use it. Staying on one volume also makes the existing cache-publish `fs::rename` a same-volume, atomic rename.

**Make the failure explain itself.** Added `PROJ-C015` ("Projection inputs and outputs do not share a filesystem root") with a `help:` line naming both offending absolute paths and the remedy, and moved the root derivation into `resolveBuildRoot()` in `packages/webui/src/projection/build-root.ts`, exported from the `@microsoft/webui/projection.js` adapter SPI alongside the other helpers adapters need (`AdapterContext`, `compileProjection`, `hashContent`).

**Let the note reach the terminal.** Press's esbuild runner swallowed diagnostic notes behind esbuild's own summary message. It now catches the failure and re-formats the raw messages. It sets `process.exitCode` rather than calling `process.exit`, because Press reads that stderr through a pipe and an immediate exit can truncate the very message this change exists to deliver.

## Verification

Reproduced and fixed end to end on a real two-drive setup (checkout on `E:`, `TEMP` at the default `C:\...\Temp`):

- Before: build fails, and with the new diagnostic it now names `C:\...\Temp\webui-press-0.0.29-.../template/docs-search/docs-search.ts` against the `E:` manifest instead of the opaque `PROJ-C013`.
- After: the same command builds cleanly (38 pages), the cache lands in `docs/.webui-press/.webui-press-cache`, and scratch directories are cleaned up.

Tests: 5 new Rust tests for volume detection and fallback (the cross-volume case is modeled with a UNC path so it runs without a second drive letter), 4 new TypeScript tests for `resolveBuildRoot` including a Windows-only assertion that the `PROJ-C015` help names both paths. `cargo xtask check` passes.

## Notes for reviewers

- `DESIGN.md` is updated with the `PROJ-C015` row, the single-root adapter contract, and Press's volume-aware cache behavior.
- Removing `commonAncestor` from the esbuild adapter also drops a pre-existing unreachable empty-array branch; the call site always passes `manifestPath` as the first element.
- The fallback directory writes its own `.gitignore` containing `*`, so consuming repositories need no Press-specific ignore entries.
- The second commit is a self-review pass that removes a needless `OnceLock` (it would have handed a second project in the same process the first project's base) and the `process.exit` truncation hazard described above.


crates/webui-press/src/build.rs
```diff
@@ -297,6 +297,15 @@ pub fn build_docs_with_cache(
     // at `<basePath>`, so the same flat layout works.
     let site_dir = out_dir.to_path_buf();
 
+    // Per-page and 404 scratch directories stay on the project's own volume so
+    // a whole Press build never spans two filesystem roots.
+    let scratch_root = crate::scratch::scratch_base(config_dir).map_err(|e| {
```
mddinbox: `scratch_base(config_dir)` chooses the cache volume from the config directory, but `outDir` is an independent path and may be absolute. If `config_dir` is on E: while `outDir` is on C: (and TEMP is on C:), this moves the embedded TypeScript and per-page scratch inputs to E: while the manifest and bundle outputs remain on C:. `resolveBuildRoot()` then correctly reports `PROJ-C015`, so a valid multi-drive configuration still fails—the new fix should select/validate the scratch base against the actual output/build volume (and the same change is needed for `extract_embedded_assets(&config_dir)` in `main.rs`).

mohamedmansour: Good catch, you're right about the anchor. I verified `out_dir` is `Path::new(&config.out_dir)`, resolved independently of the config file, so it can be absolute or working-directory relative and land on another volume.

Fixed in 5b3f8af7. `scratch_base` now takes the output directory as the volume anchor and keeps the config directory only as the home for the fallback cache, applied to both `extract_embedded_assets` in `main.rs` and the per-page/404 scratch in `build.rs`. The output directory is the right anchor because it unconditionally holds the generated entry points, the esbuild `outbase`/`outdir`, and the projection manifest, so the build root always contains it; the config directory contributes inputs only when sources happen to live there. That also fixes the sub-case you'd expect to work: output directory on the system temp volume now keeps the shared cache instead of being forced onto the config volume.

One nuance worth recording. I reproduced your exact scenario (config dir on `E:`, absolute `outDir` on `C:`, `TEMP` on `C:`) against the fix. The embedded cache is no longer the offender, but the build still fails, now pointing at the real pair:

```
help: "C:\...\webui-press-outvol-dist\webui-projection.json" and
      "E:\...\packages\webui-framework\dist\template-events.js"
      are on different filesystem roots ...
```

When a project's own sources are on `E:` and its configured output is on `C:`, there is no expressible build root and no scratch placement can create one, since both ends are fixed by the user's configuration. So that configuration is genuinely unbuildable rather than something the cache was breaking; what changed is that `PROJ-C015` now names the actual conflicting pair instead of misattributing it to Press's temp cache. I also broadened the help text to name a configured output directory alongside `TEMP`/`TMP`.

Added `output_volume_decides_the_base_not_the_config_directory` to cover the new rule. Full gate passes.

mddinbox (approved):
Reviewed the current head 5b3f8af7 and rechecked the prior cross-volume feedback. The scratch/cache anchor now follows the configured output volume, the projection root validation reports genuine cross-root configurations, and the related Press tests plus green CI provide sufficient validation. No remaining issues found.

